PR #1, opened 2025-10-14
Add Deno/ESM compatibility using tsup bundler

a-bro:
## Summary

  Converts the build system from vanilla tsc to tsup to generate proper ES
  modules with explicit file extensions, enabling the package to work
  seamlessly in Deno, modern Node.js, and browsers while maintaining
  backward compatibility.

  ## Problem

  The package was built with TypeScript's tsc compiler using CommonJS
  output ("module": "commonjs"), which caused issues when importing in
  Deno:
  - Deno requires explicit .js extensions in ES module imports
  - tsc doesn't add file extensions to emitted imports (e.g., export * from
   './types' instead of export * from './types.js')
  - CommonJS format is incompatible with Deno's native ESM-only runtime

  ## Solution

  Replaced the build pipeline with tsup, a modern bundler designed for
  TypeScript libraries that:
  - Automatically generates proper ESM with correct import paths
  - Supports dual-format output (ESM + CommonJS) from a single build
  - Produces optimized, tree-shakeable bundles
  - Handles all module resolution edge cases

  ## Changes

  Build System

  - Added: tsup as dev dependency
  - Added: tsup.config.ts configuration file
  - Updated: package.json build scripts to use tsup instead of tsc
  - Updated: tsconfig.json to "module": "ESNext" and "moduleResolution": 
  "bundler"

  ## Package Configuration

  - Added: "type": "module" to declare package as ESM
  - Added: "exports" field with proper import/require mappings:
    - import → dist/index.js (ESM)
    - require → dist/index.cjs (CommonJS)
    - types → dist/index.d.ts
  - Updated: "main" to point to dist/index.cjs (CommonJS fallback)
  - Updated: "module" to point to dist/index.js (ESM)
  - Bumped: Version from 0.1.0 → 0.1.2

  ## Output

  Build now generates:
  - dist/index.js - ES module bundle (Deno, modern Node.js, browsers)
  - dist/index.cjs - CommonJS bundle (legacy Node.js)
  - dist/index.d.ts - TypeScript definitions for ESM
  - dist/index.d.cts - TypeScript definitions for CommonJS
  - Sourcemaps for both formats

  ## Compatibility

  ✅ Deno - Full support via proper ESM format✅ Node.js (ESM) - Modern
  Node.js using import✅ Node.js (CJS) - Legacy Node.js using require()✅
  Browsers - Via bundlers (Webpack, Vite, etc.)✅ TypeScript - Full type
  support for all environments

  ## Testing

  Tested in Supabase Edge Functions (Deno runtime) with successful import
  and execution of getSummaryAndTags() function.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mux/ai",
-  "version": "0.1.1",
+  "version": "0.1.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@mux/ai",
-      "version": "0.1.1",
+      "version": "0.1.2",
       "license": "MIT",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.28.0",
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@mux/ai",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@mux/ai",
-      "version": "0.1.0",
+      "version": "0.1.1",
       "license": "MIT",
       "dependencies": {
         "@anthropic-ai/sdk": "^0.28.0",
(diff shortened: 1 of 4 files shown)


philcluff:
Ignoring copilot's latest suggestions as they seem to conflict with the current approach (removing type module)

cjpillsbury (approved):
This all looks very reasonable to me. Bigger picture, we can probably make es modules the default/main for modern envs these days (including node), but this is at least a step in the right direction! Approved!

PR #2, opened 2025-11-18
First pass of refactoring and cleanup

philcluff:
First pass of cleaning up the vibe coded first pass of the @mux/ai repo.

This PR is mostly simple cleanup of trying to make sure we have code reuse wherever we should, and that the examples and config are consistent and well documented. 

AI generated summary:
  - Created src/lib/client-factory.ts - centralized client initialization and credential validation
  - Created src/lib/provider-models.ts - centralized provider configuration and defaults
  - Created src/lib/retry.ts - unified retry logic with exponential backoff
  - Refactored chapters.ts, summarization.ts, moderation.ts to eliminate ~300 lines of duplication
  - Added npm scripts to package.json for running all examples from root
  - Standardized all examples (17 files) to use import 'dotenv/config'
  - Fixed example script usage messages to match npm script names
  - Updated README with CLI usage section and prerequisites
  - Net reduction: 87 lines (-375 deletions, +462 insertions from new abstractions)


src/lib/client-factory.ts
```diff
@@ -0,0 +1,123 @@
...
+  openaiApiKey?: string;
+  anthropicApiKey?: string;
+}
+
+/**
+ * Validates and retrieves credentials from options or environment variables
+ */
+export function validateCredentials(
+  options: ClientCredentials,
+  requiredProviders?: Array<'openai' | 'anthropic'>
```
monsieurBoutte: 🔧 one small thing here. The `requiredProviders?: Array<'openai' | 'anthropic'>` type reads like we support multiple providers at once, but the actual usage doesn’t really line up with that.

The only call site wraps a single provider into an array `(provider ? [provider] : undefined)`, which makes it feel like each workflow is only meant to use one provider at a time.

src/lib/client-factory.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Creates a Mux client with validated credentials
+ */
+export function createMuxClient(credentials: ValidatedCredentials): Mux {
+  return new Mux({
+    tokenId: credentials.muxTokenId,
+    tokenSecret: credentials.muxTokenSecret,
+  });
+}
```
monsieurBoutte, line 69: 🔧 Here in `createMuxClient` we assume `ValidatedCredentials` already has `muxTokenId` and `muxTokenSecret`, but the OpenAI and Anthropic wrappers both double check their own required keys.

Feels like it might be worth adding the same defensive check in this wrapper too. It keeps the behavior consistent across providers and keeps each wrapper responsible for its own invariants. Especially if we're going to export these functions out of this file

README.md
```diff
@@ -549,7 +549,67 @@ const result = await getSummaryAndTags(
 
 ## Examples
 
-See the `examples/` directory for complete working examples:
+See the `examples/` directory for complete working examples.
```
cjpillsbury, line 552: sugestion(non-blocking): Probably room for improvement on this readme in general, though this is 👍. Some non-blocking minor callouts for this section:
- Given the length of this doc, it's probably good to assume folks may not have read everything that came before, so re-linking to things like "pre-requisites"/"setup" may be good here, presuming what's at the bottom isn't exhaustive.
- Some defs/links to what things like (mux) "asset-id", (BCP-47, I presume) language-code, and (LLM/"AI API") provider would be good.

README.md
```diff
@@ -549,7 +549,67 @@ const result = await getSummaryAndTags(
...
+npm run example:translation <asset-id> [from-lang] [to-lang]
+
+# Audio Translation (Dubbing)
+npm run example:audio-translation <asset-id> [to-lang]
+```
+
+**Examples:**
+```bash
+# Generate chapters with OpenAI
+npm run example:chapters abc123 en openai
```
cjpillsbury, line 596: suggestion(non-blocking): Given that the provider strings have semantic meaning, is there a documented list of what those are (at least by default install) in the readme? If not, they should probably be added. If so, linking to them somewhere in this section would be 👌

README.md
```diff
@@ -549,7 +549,67 @@ const result = await getSummaryAndTags(
...
+npm run example:moderation abc123
+
+# Translate captions from English to Spanish
+npm run example:translation abc123 en es
+
+# Create AI-dubbed audio in French
+npm run example:audio-translation abc123 fr
+```
+
+**Prerequisites:**
```
cjpillsbury: nitpick(non-blocking): These should probably come before the examples in the quick start. Many folks will likely try (and fail) to run the examples without having read this far.

examples/audio-translation/basic-example.ts
```diff
@@ -1,18 +1,15 @@
+import 'dotenv/config';
 import { translateAudio } from '@mux/ai';
-import { config } from 'dotenv';
 
-// Load environment variables from parent directory - override existing
-const result = config({ path: '../../.env', override: true });
-console.log('Dotenv result:', result.error ? result.error.message : 'SUCCESS');
 
 async function main() {
   const assetId = process.argv[2];
```
cjpillsbury, line 6: suggestion(non-blocking): use yargs or commander lib for this kinda crud (See, e.g. https://github.com/lirantal/nodejs-cli-apps-best-practices?tab=readme-ov-file#11-respect-posix-args)

examples/audio-translation/basic-example.ts
```diff
@@ -1,18 +1,15 @@
+import 'dotenv/config';
```
cjpillsbury, line 1: praise: This is good. I like that there's significant separation of the CLI vs. the functional logic, and having this "baked in" to the CLI side of that feels like the right level of seams.

src/lib/client-factory.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Factory for creating all necessary clients for a workflow
+ */
+export interface WorkflowClients {
+  mux: Mux;
+  openai?: OpenAI;
+  anthropic?: Anthropic;
+  credentials: ValidatedCredentials;
+}
```
monsieurBoutte, line 103: 💭 One type-safety thought here. `WorkflowClients` is currently typed with providers being optional, so callers lose the guarantee that “if provider is X, then client X definitely exists.”

One way to tighten this up would be to make `WorkflowClients` provider-aware, for example with a `WorkflowClientsFor<P>` mapped type and a generic
`createWorkflowClients<P extends Provider>(options, provider: P): WorkflowClientsFor<P>`.

```ts
type Provider = 'openai' | 'anthropic' | undefined;

type WorkflowClientsFor<P extends Provider> =
  P extends 'openai'
    ? {
        mux: Mux;
        openai: OpenAI;          // required
        anthropic?: Anthropic;
        credentials: ValidatedCredentials;
      }
  : P extends 'anthropic'
    ? {
        mux: Mux;
        openai?: OpenAI;
        anthropic: Anthropic;    // required
        credentials: ValidatedCredentials;
      }
  : {
      mux: Mux;
      openai?: OpenAI;
      anthropic?: Anthropic;
      credentials: ValidatedCredentials;
    };

// Then type createWorkflowClients generically
export function createWorkflowClients<P extends Provider>(
  options: MuxAIOptions,
  provider: P
): WorkflowClientsFor<P> {
  // impl as today, just typed as WorkflowClientsFor<P>
}
```

Another option would be provider-specific helpers that return a shape where the corresponding client is required.

Not blocking at all, but it feels like a nice future improvement for callers, especially if we're integrating directly with provider sdks.

examples/audio-translation/basic-example.ts
cjpillsbury, line 1: idea(non-blocking): Although node and other runtimes have direct access to process.env, you may want to consider centralizing that to the CLI side and passing relevant config down to the functional abstractions as a followup. That just removes more assumptions. I generally try to avoid bleeding "ambient environment dependencies" too deeply into my code if/when possible, and this architecture is already aiming in a direction that lends itself to that.

src/lib/provider-models.ts
```diff
@@ -0,0 +1,37 @@
...
+  default: string;
+  fast: string;
+  powerful: string;
+}
+
+export const PROVIDER_MODELS: Record<string, ProviderModelConfig> = {
+  openai: {
+    default: 'gpt-4o-mini',
+    fast: 'gpt-4o-mini',
+    powerful: 'gpt-4o',
```
monsieurBoutte: 🔧 I'd recommend us defaulting to newer [gpt-5 variants](https://platform.openai.com/docs/models/compare?model=gpt-5.1) with `gpt-5.1` as the default, `gpt-5-mini` as fast, and `gpt-5-pro` as powerful

src/lib/provider-models.ts
```diff
@@ -0,0 +1,37 @@
...
+  },
+  anthropic: {
+    default: 'claude-3-5-haiku-20241022',
+    fast: 'claude-3-5-haiku-20241022',
+    powerful: 'claude-3-5-sonnet-20241022',
+  },
+};
```
monsieurBoutte, line 21: 🔧  similarly here, I'd recommend us defaulting to the [4.5 model family](https://docs.claude.com/en/docs/about-claude/models/overview)

src/burned-in-captions.ts
```diff
@@ -32,7 +32,7 @@ const burnedInCaptionsSchema = z.object({
   detectedLanguage: z.string().nullable()
 });
 
-const DEFAULT_SYSTEM_PROMPT = `You are an expert at analyzing video frames to detect burned-in captions (also called open captions or hardcoded subtitles). These are text overlays that are permanently embedded in the video image, common on TikTok, Instagram Reels, and other social media platforms.
+const OPENAI_SYSTEM_PROMPT = `You are an expert at analyzing video frames to detect burned-in captions (also called open captions or hardcoded subtitles). These are text overlays that are permanently embedded in the video image, common on TikTok, Instagram Reels, and other social media platforms.
```
cjpillsbury, line 35: thought(non-blocking): I wonder if these strings should be pulled out of the inline code. Definitely not needed now, but food for thought/something to have on the radar for future iterations.

monsieurBoutte (approved):
👍 no blocking comments but left behind some well intentioned thoughts to consider

src/chapters.ts
```diff
@@ -227,13 +185,17 @@ export async function generateChapters(
   let chaptersData: { chapters: Chapter[] } | null = null;
 
   if (provider === 'openai') {
+    if (!clients.openai) {
```
cjpillsbury, line 188: thought(non-blocking): these feel like further places where some architectural seams/abstractions would be good. I think a bigger longer term "north star" would be to have modules like this not have to "know" about openai or anthropic at all. Looking at this code, that feels like a plausible, possible future.

src/lib/client-factory.ts
```diff
@@ -0,0 +1,123 @@
+import Mux from '@mux/mux-node';
```
cjpillsbury, line 1: praise: This is definitely a step in the right direction. Like the callout above, longer term, a "client factory" probably shouldn't know about specific 3rd party APIs but should instead have generic abstractions where particular steps can be "pulled out" for API-specific crud.

src/lib/provider-models.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * Default model configurations for each provider
+ */
+export interface ProviderModelConfig {
+  default: string;
+  fast: string;
+  powerful: string;
+}
+
+export const PROVIDER_MODELS: Record<string, ProviderModelConfig> = {
```
cjpillsbury, line 10: thought(non-blocking): This should be JSON config that is overridable. The more I'm seeing, I'm thinking we may want a well defined ai config object that other more robust CLIs have (maybe replacing dotenv, which is more for servers/services/etc). Happy to have followup conversations here.

src/lib/retry.ts
```diff
@@ -0,0 +1,65 @@
...
+  const delayWithJitter = exponentialDelay * (0.5 + Math.random() * 0.5);
+  return Math.min(delayWithJitter, maxDelay);
+}
+
+/**
+ * Executes an async function with retry logic
+ */
+export async function withRetry<T>(
+  fn: () => Promise<T>,
+  options: RetryOptions = {}
```
cjpillsbury: nitpick(non-blocking): consider using defaults and destructuring in the param for consistency.

cjpillsbury (approved):
TL;DR - These all look like steps in the right direction, and there are good candidates for additional steps in said direction (mentioned in review comments). Approved.

src/burned-in-captions.ts
philcluff: Yeh, I had the same thought! I'm going to leave that as-is for now and re-review when we start to explore switching our for the vercel ai abstraction I think.

README.md
philcluff: Great call out! I'm getting super conscious of how long this is, but want to retain the full API documentation - do you know any good projects with long, but well navigable readmes, or should I be breaking this down into multiple, or something else?

examples/audio-translation/basic-example.ts
philcluff: Yes, agreed, I'm going to user commander, but I'm going to do it in a separate PR, good shout.

src/lib/client-factory.ts
philcluff: Good catch! Added!

src/lib/provider-models.ts
philcluff: Yep, good spot, this was nice and easy and worked well!

src/lib/provider-models.ts
philcluff: Yes, good call...

BUT this broke functionality, because the newer models behave differently when it comes to outputting JSON. 

I'm eager that we get moved over to structured outputs on anthropic ASAP, so I might have to delay this one annoyingly.

src/lib/client-factory.ts
philcluff: Great spot - addressed.

src/lib/client-factory.ts
```diff
@@ -0,0 +1,123 @@
...
+  const clients: WorkflowClients = {
+    mux: createMuxClient(credentials),
+    credentials,
+  };
+
+  if (provider === 'openai' || credentials.openaiApiKey) {
+    clients.openai = createOpenAIClient(credentials);
+  }
+
+  if (provider === 'anthropic' || credentials.anthropicApiKey) {
```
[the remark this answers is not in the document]
philcluff: Done

src/lib/retry.ts
philcluff: Good spot. Done.

src/lib/client-factory.ts
philcluff: Yeh, great feedback. 

I'm not going to do this one right now because I want to feel this out more once we have an abstraction library in place, I think we might feel differently then.

src/lib/client-factory.ts
philcluff: Yep agreed - tabling this one for now and will come back to it when we are exploring @vercel/ai et-al.

src/chapters.ts
philcluff: Yep, agreed - again will revisit when we look at abstractions for vendor specific SDKs

README.md
philcluff: Hitting resolve here to merge - will pick up bigger improvements on README in the future.

README.md
philcluff: Done!

src/lib/provider-models.ts
philcluff: Yep - I'll come back to this one after we explore abstractions.

PR #5, opened 2025-11-21
chore: ai-sdk integration

monsieurBoutte:
## Branch Overview

This branch clarifies the public surface into low-level `primitives` and higher-level `functions`, modernizes provider integration via the Vercel AI SDK, and refreshes examples/docs to match.

### API Surface & Package Boundaries
- **Primitives (`@mux/ai/primitives`)**: small, composable helpers for working with Mux assets (thumbnails, transcripts, storyboards, moderation, etc.), returning simple data structures and leaving orchestration to the caller.
- **Functions (`@mux/ai/functions`)**: opinionated, end-to-end workflows (summarization, chapters, translation, audio translation, burned-in captions) that compose primitives, Mux APIs, and LLM providers for common use cases.
- `src/index.ts` now re-exports the shallow `primitives` and `functions` modules so users can import from `@mux/ai`, `@mux/ai/functions`, or `@mux/ai/primitives` explicitly.
- `tsconfig` path aliases and `tsup`/`package.json` exports align the source layout with the published entry points, and we now build standalone bundles for the root index plus the `functions` and `primitives` packages.

### Provider Plumbing
- Added `src/lib/providers.ts` and expanded `src/lib/client-factory.ts` to resolve OpenAI, Anthropic, and Google LLMs through the Vercel AI SDK.
- Provider defaults, model IDs, and credential validation are centralized, including support for Google (`gemini-2.5-flash`) alongside updated OpenAI (`gpt-5-mini`) and Anthropic models.
- Introduced shared helpers (`src/lib/mux-assets.ts`, transcript/thumbnail/storyboard utilities) for asset fetching and text-track parsing across workflows.

### Provider Strategy & Entry Points
- The new provider abstraction makes higher-level `functions` effectively provider-agnostic, with consistent options across OpenAI, Anthropic, and Google.
- Today, users integrate via a **"bring your own keys"** model, passing provider API keys explicitly or via environment variables.
- The design also reserves a parallel, future **"use Mux credits"** path, where the same functions can be backed by Mux-hosted models without changing call sites.

### Feature Updates
- Functions now import shared helpers, rely on the unified provider resolver, and expose richer metadata (titles, descriptions, tags, storyboard URLs, etc.).
- Moderation primitives include storyboard/thumbnail helpers and multi-provider support (OpenAI, Anthropic, Google) with configurable thresholds.
- Transcript parsing logic is consolidated in `src/primitives/transcripts.ts` and reused by higher-level workflows like summarization and chapters.

### Examples & Documentation
- Examples now use the updated module paths (`@mux/ai/functions`, `@mux/ai/primitives`) and demonstrate provider selection, updated CLI usage, and new model defaults.
- README files document credential requirements, provider selection, and return shapes in terms of the new `primitives` vs `functions` split.
- Example TypeScript configs ensure samples compile cleanly against the shared sources and path aliases.


src/functions/translation.ts
```diff
@@ -1,90 +1,57 @@
...
+    s3SecretAccessKey: providedS3SecretAccessKey,
+    uploadToMux: uploadToMuxOption,
+    ...clientConfig
   } = options;
 
-  if (provider !== 'anthropic') {
-    throw new Error('Only Anthropic provider is currently supported for translation');
+  if (!provider) {
+    throw new Error('Translation provider is required. Set options.provider to one of the supported providers (openai, anthropic, google).');
   }
```
mihar-22: Noticed that the other workflows default to `openai` but this one errors out. Curious, any reason we're inconsistent? Since `openai` is supported here as well based on the message, should we default to it as well and then only error out if an unsupported provider is given?

README.md
```diff
@@ -6,23 +6,53 @@ AI-powered video analysis library for Mux, built in TypeScript.
...
-| `translateCaptions` | Translate video captions to different languages | Anthropic only | `claude-sonnet-4-20250514` | Asset ID + languages + S3 config | Translated VTT + Mux track ID |
+| `getSummaryAndTags` | Generate titles, descriptions, and tags from a Mux video asset | OpenAI, Anthropic, Google | `gpt-5-mini`, `claude-sonnet-4-5`, `gemini-2.5-flash` | Asset ID + options | Title, description, tags, storyboard URL |
+| `getModerationScores` | Analyze video thumbnails for inappropriate content | OpenAI, Anthropic, Google | `omni-moderation-latest` (OpenAI) or AI SDK defaults (`gpt-5-mini`, `claude-sonnet-4-5`, `gemini-2.5-flash`) | Asset ID + thresholds | Sexual/violence scores, flagged status |
+| `hasBurnedInCaptions` | Detect burned-in captions (hardcoded subtitles) in video frames | OpenAI, Anthropic, Google | `gpt-5-mini`, `claude-sonnet-4-5`, `gemini-2.5-flash` | Asset ID + options | Boolean result, confidence, language |
+| `generateChapters` | Generate AI-powered chapter markers from video captions | OpenAI, Anthropic, Google | `gpt-5-mini`, `claude-sonnet-4-5`, `gemini-2.5-flash` | Asset ID + language + options | Timestamped chapter list |
```
mihar-22, line 12: Small callout but if we have a consistent set of default models for each provider then maybe we can just list them out at the top and then only callout when it's not the case. Otherwise table is quite noisy. If this is not the intended home for the official docs then feel free to ignore.

src/functions/summarization.ts
```diff
@@ -0,0 +1,159 @@
...
+import { downloadImageAsBase64 } from '../lib/image-download';
+import { createWorkflowClients } from '../lib/client-factory';
+import { withRetry } from '../lib/retry';
+import { SupportedProvider } from '../lib/providers';
+import { fetchPlaybackAsset } from '../lib/mux-assets';
+import { fetchTranscriptForAsset } from '../primitives/transcripts';
+import { getStoryboardUrl } from '../primitives/storyboards';
+
+const DEFAULT_PROMPT =
+  "Generate a short title (max 100 characters) and description (max 500 characters) for what happens. Start immediately with the action or subject - never reference that this is a video, content, or storyboard. Provide up to 10 concise keywords that capture the primary people, objects, or actions. Example: Title: 'Cooking Pasta Tutorial' Description: 'Someone cooks pasta by boiling water and adding noodles.' Keywords: ['cooking', 'pasta', 'boiling water', 'noodles', 'kitchen'].";
```
mihar-22, line 63: Not a bad idea to export prompts so devs can build/compose their own calls?

src/types.ts
```diff
@@ -1,13 +1,291 @@
...
+  /** Longer description of the detected content. */
+  description: string;
+  /** Up to 10 keywords extracted by the model. */
+  tags: string[];
+  /** Storyboard image URL that was analyzed. */
+  storyboardUrl: string;
+}
+
+/** Configuration accepted by `getSummaryAndTags`. */
+export interface SummarizationOptions extends MuxAIOptions {
```
mihar-22: Small nit but I feel like some of these types should live alongside their implementation. Felt a little strange jumping to a shared types file to find this.

src/functions/summarization.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+  }
+
+  return normalized;
+}
+
+export async function getSummaryAndTags(
+  assetId: string,
+  promptOrOptions?: string | SummarizationOptions,
+  options?: SummarizationOptions
```
mihar-22, line 100: Our options don't explicitly allow `AbortSignal` to be passed down to all fetch calls and the AI call which will mean execution keeps running ($$). This is true for all workflows.

One step further, it would be really nice to also override [`generateObject` parameters](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-object#parameters). Exposing a config option by default for all workflows would be awesome (e.g., `ai`?). Again, this would be true for all workflows.

src/functions/summarization.ts
```diff
@@ -0,0 +1,159 @@
...
+  if (transcriptText) {
+    const transcriptType = cleanTranscript ? 'transcript' : 'WebVTT transcript';
+    contextualPrompt += ` Use the following ${transcriptType} for additional context: "${transcriptText}"`;
+  }
+
+  // Analyze storyboard with AI provider
+  const imageUrl = getStoryboardUrl(playbackId, 640);
+
+  const analyzeStoryboard = async (imageDataUrl: string) => {
+    const response = await generateObject({
```
mihar-22, line 163: What if we want to [stream](https://ai-sdk.dev/docs/reference/ai-sdk-core/stream-object) the result client-side? For peak UX no one will be waiting `n` seconds for a response to finish. Could this be a possibility for any of our workflows? Are we boxed in here? 

One idea is providing a option to "generate" or "stream" and then using inferred conditional types.

src/functions/summarization.ts
```diff
@@ -0,0 +1,159 @@
...
+          role: 'user',
+          content: [
+            { type: 'text', text: contextualPrompt },
+            { type: 'image', image: imageDataUrl },
+          ],
+        },
+      ],
+    });
+
+    return response.object;
```
mihar-22, line 183: We're removing a lot of helpful information from the [response](https://ai-sdk.dev/docs/reference/ai-sdk-core/generate-object#returns), things like `finishReason` and more importantly `usage`. I think we should still include the `response` object in the return. Same for all workflows.

mihar-22 (commented):
I think the refactor is a massive step in the right direction. So many decisions I love such as the client factory using Vercel AI SDK under the hood (cleaning up and improving provider plumbing), separation of "primitives" and "functions", and exporting schema/types ❤️ **Amazing work, Victor!**

Eventually when this has docs outside of a readme, it'd be nice to see an "ejected" example of each workflow where we can provide a custom prompt with primitives, exported schema, overrides, etc.. The underlying prompt being exported but also surfaced in the docs would be great!

Curious if we could/should have a Next.js export eventually with [caching](https://nextjs.org/docs/app/getting-started/caching-and-revalidating) handled. Worth a discussion at some point?

src/functions/translation.ts
monsieurBoutte: I have a follow-up PR that addresses this issue for Anthropic. I believe it was a downstream problem at the Zod schema level, where it attempted to set maximum character lengths for titles and maximum word counts for tags. I've encountered similar oddities with Zod schemas and LLM outputs before, and I usually resolve them with some prompt adjustments rather than imposing strict limits at the schema level.

src/functions/summarization.ts
monsieurBoutte: Agreed, in my follow up PR I'm going to propose a prompt builder pattern to help support this idea of more malleable inputs that lend themselves well to better context engineering optionality at the call site for users

src/functions/summarization.ts
monsieurBoutte: 💡 such a good call out! I'll be sure to add support for AbortSignals

src/functions/summarization.ts
monsieurBoutte: yeah for sure, another solid call out. I'll hold off on adding that within the scope of this PR but I'm down to follow up with that support in a separate PR

README.md
monsieurBoutte: I'd like to follow up with a PR to tighten the brevity of our README. Ideally, we keep it straightforward for developers to understand without wading through too many extraneous details.

mihar-22: ➕

examples/burned-in-captions/provider-comparison.ts
```diff
@@ -14,49 +14,54 @@ async function main() {
...
-    const openaiStart = Date.now();
-    const openaiResult = await hasBurnedInCaptions(assetId, {
-      provider: 'openai'
-    });
-    const openaiDuration = Date.now() - openaiStart;
+    type ProviderConfig = { name: string; provider: 'openai' | 'anthropic' | 'google' };
+    const providers: ProviderConfig[] = [
+      { name: 'OpenAI', provider: 'openai' },
+      { name: 'Anthropic', provider: 'anthropic' },
+      { name: 'Google', provider: 'google' },
```
philcluff, line 21: Future improvement: We should think about this abstraction a little more carefully at some stage - for example, should this name answer "Is this Vertex, AI Studio, or some other Google way of running AI tooling" ?

examples/chapters/basic-example.ts
```diff
@@ -1,11 +1,11 @@
 import 'dotenv/config';
-import { generateChapters } from '../../src/chapters';
+import { generateChapters } from '@mux/ai/functions';
 
 async function main() {
   const assetId = process.argv[2];
   const languageCode = process.argv[3] || 'en';
-  const provider = process.argv[4] as 'openai' | 'anthropic' || 'openai';
-  
+  const provider = (process.argv[4] as 'openai' | 'anthropic' | 'google') || 'openai';
```
philcluff, line 7: Future: Next work item after this and tests are merged needs to be CLI parsing

examples/moderation/basic-example.ts
```diff
@@ -1,38 +1,67 @@
...
 
+  const provider = providerArg;
+  const model = provider === 'hive' ? undefined : DEFAULT_MODELS[provider];
+
   // Debug: Check if env vars are loaded
   console.log('Debug - Environment variables:');
-  console.log('MUX_TOKEN_ID:', process.env.MUX_TOKEN_ID ? `${process.env.MUX_TOKEN_ID.substring(0, 10)}...` : 'NOT SET');
-  console.log('MUX_TOKEN_SECRET:', process.env.MUX_TOKEN_SECRET ? `${process.env.MUX_TOKEN_SECRET.substring(0, 10)}...` : 'NOT SET'); 
-  console.log('OPENAI_API_KEY:', process.env.OPENAI_API_KEY ? `${process.env.OPENAI_API_KEY.substring(0, 10)}...` : 'NOT SET');
+  console.log('MUX_TOKEN_ID:', mask(process.env.MUX_TOKEN_ID));
```
philcluff: Nit: This was debugging and can be stripped out

examples/moderation/basic-example.ts
```diff
@@ -1,38 +1,67 @@
 import 'dotenv/config';
-import { getModerationScores } from '@mux/ai';
+import { getModerationScores } from '@mux/ai/primitives';
```
philcluff: I'm 50/50 on if this should be primitive or function. 

Right now, this is a function. 

In the future, we want to make it a primitive. 

But given the current implementation, I think I'd push on changing it back to a function until it's more native.

examples/moderation/google-example.ts
```diff
@@ -0,0 +1,48 @@
+import 'dotenv/config';
```
philcluff: Does basic example with a given provider not just do the same thing? Could we just merge these examples to the basic one with the appropriate argument?

examples/moderation/hive-example.ts
```diff
@@ -1,47 +1,73 @@
 import 'dotenv/config';
-import { getModerationScores } from '@mux/ai';
+import { getModerationScores } from '@mux/ai/primitives';
 
+const usage = `
```
philcluff: Same as above - maybe just merge these all together.

examples/moderation/provider-comparison.ts
```diff
@@ -1,58 +1,86 @@
...
-      getModerationScores(assetId, {
-        provider: 'hive',
-        thresholds: { sexual: 0.7, violence: 0.8 }
+    const configs = [
+      {
+        label: 'OpenAI',
+        options: {
+          provider: 'openai' as const,
+          model: 'omni-moderation-latest',
+          thresholds: { sexual: 0.7, violence: 0.8 },
```
philcluff: Do thresholds have a default we could use instead?

examples/summarization/anthropic-example.ts
```diff
@@ -13,7 +12,7 @@ async function main() {
   // Debug: Check if env vars are loaded
   console.log('Debug - Environment variables:');
```
philcluff: We might want to strip all these out now we're not debugging creds too much

src/functions/translate-audio.ts
```diff
@@ -4,27 +4,101 @@ import { S3Client } from '@aws-sdk/client-s3';
 import { Upload } from '@aws-sdk/lib-storage';
 import { GetObjectCommand } from '@aws-sdk/client-s3';
 import { getSignedUrl } from '@aws-sdk/s3-request-presigner';
-import { MuxAIOptions } from './types';
+import { AudioTranslationOptions, AudioTranslationResult } from '../types';
 
-export interface AudioTranslationResult {
+const STATIC_RENDITION_POLL_INTERVAL_MS = 5000;
```
philcluff, line 50: I hate including polling Mux, but yeh, it's the only solution right now

src/functions/summarization.ts
```diff
@@ -0,0 +1,159 @@
...
+  if (typeof promptOrOptions === 'string') {
+    prompt = promptOrOptions;
+    actualOptions = options || {};
+  } else {
+    prompt = DEFAULT_PROMPT;
+    actualOptions = promptOrOptions || {};
+  }
+
+  const {
+    provider = 'openai',
```
philcluff: Should our default providers be consts at the top of the file like prompts are for each tool?

src/functions/summarization.ts
```diff
@@ -0,0 +1,159 @@
...
+
+  const analyzeStoryboard = async (imageDataUrl: string) => {
+    const response = await generateObject({
+      model: clients.languageModel.model,
+      schema: summarySchema,
+      messages: [
+        {
+          role: 'system',
+          content:
+            'You are an image analysis tool. You will be given a storyboard image from a video showing multiple frames/scenes, and be expected to return structured data about the contents across all the frames.',
```
philcluff: Future improvement: We should seed the LLM with context on the temporal sequencing of these frames - the storyboard has tiles in x and y, so temporal ordering is isn't explicit.

src/functions/summarization.ts
```diff
@@ -0,0 +1,159 @@
...
+    }
+  } catch (error: unknown) {
+    throw new Error(
+      `Failed to analyze video content with ${provider}: ${error instanceof Error ? error.message : 'Unknown error'}`
+    );
+  }
+
+  return {
+    assetId,
+    title: aiAnalysis.title || 'No title available',
```
philcluff: S1: Either throw or return empty string if we can't generate anything, or customers could easily end up with this text on their assets / site.

src/functions/translation.ts
```diff
@@ -1,90 +1,57 @@
...
+    s3AccessKeyId: providedS3AccessKeyId,
+    s3SecretAccessKey: providedS3SecretAccessKey,
+    uploadToMux: uploadToMuxOption,
+    ...clientConfig
   } = options;
 
-  if (provider !== 'anthropic') {
-    throw new Error('Only Anthropic provider is currently supported for translation');
+  if (!provider) {
+    throw new Error('Translation provider is required. Set options.provider to one of the supported providers (openai, anthropic, google).');
```
philcluff: Future: We need to test this one specifically with much longer / larger assets than we have today, as I'm sure this will error out at some duration.

src/functions/translate-captions.ts
```diff
@@ -1,90 +1,57 @@
...
-  s3SecretAccessKey?: string;
-  uploadToMux?: boolean;
-}
+import { generateObject } from 'ai';
+import { TranslationResult, TranslationOptions, translationSchema } from '../types';
+import { createWorkflowClients } from '../lib/client-factory';
+import { SupportedProvider } from '../lib/providers';
 
-export async function translateCaptions(
+export async function translateCaptions<P extends SupportedProvider = SupportedProvider>(
```
philcluff, line 63: Can you rename this file to "captions-translation.ts` or alike?

src/lib/client-factory.ts
```diff
@@ -1,33 +1,42 @@
...
-  requiredProvider?: 'openai' | 'anthropic'
+  requiredProvider?: SupportedProvider
 ): ValidatedCredentials {
   const muxTokenId = options.muxTokenId || process.env.MUX_TOKEN_ID;
   const muxTokenSecret = options.muxTokenSecret || process.env.MUX_TOKEN_SECRET;
   const openaiApiKey = options.openaiApiKey || process.env.OPENAI_API_KEY;
   const anthropicApiKey = options.anthropicApiKey || process.env.ANTHROPIC_API_KEY;
+  const googleApiKey =
+    options.googleApiKey ||
+    process.env.GOOGLE_GENERATIVE_AI_API_KEY ||
```
philcluff: Why can this be one of two different names? Is that normal in the Google AI ecosystem?

src/lib/mux-assets.ts
```diff
@@ -0,0 +1,16 @@
+import Mux from '@mux/mux-node';
+import { MuxAsset, PlaybackAsset } from '../types';
+
+function ensurePlaybackId(asset: MuxAsset): string {
```
philcluff: Rename to `ensurePublicPlaybackId`

src/lib/mux-assets.ts
```diff
@@ -0,0 +1,16 @@
...
+  const playbackId = asset.playback_ids?.[0]?.id;
```
philcluff: Check if it's actually public (usable)

src/lib/mux-assets.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (!playbackId) {
+    throw new Error('No playback ID found for this asset');
+  }
+  return playbackId;
+}
+
+export async function fetchPlaybackAsset(mux: Mux, assetId: string): Promise<PlaybackAsset> {
```
philcluff: I don't like this function name, it doesn't match a concept in Mux Video.

src/lib/providers.ts
```diff
@@ -0,0 +1,109 @@
+import { createAnthropic } from '@ai-sdk/anthropic';
+import { createGoogleGenerativeAI } from '@ai-sdk/google';
+import { createOpenAI } from '@ai-sdk/openai';
+import { LanguageModel } from 'ai';
+import { MuxAIOptions } from '../types';
+
+// Hosted Mux Gemini path intentionally not exposed yet; limited to BYO providers for now.
```
philcluff: Remove this mention of Mux hosted Gemini

src/lib/providers.ts
```diff
@@ -0,0 +1,109 @@
...
+  model?: ModelIdByProvider[P];
+}
+
+export interface ResolvedModel<P extends SupportedProvider = SupportedProvider> {
+  provider: P;
+  modelId: ModelIdByProvider[P];
+  model: LanguageModel;
+  /**
+   * Indicates whether the model relies on Mux tokens instead of BYO provider keys.
+   * (Future hosted Mux models will flip this to true.)
```
philcluff: Same

src/lib/providers.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (!value) {
+    throw new Error(`Missing ${name}. Set ${name} in your environment or pass it in options.`);
+  }
+  return value;
+}
+
+/**
+ * Resolves a language model using Vercel's AI SDK providers.
+ * - BYO: defaults to the provider-specific env var conventions used by ai-sdk
+ * - Mux-hosted: reserved for a future path that does not require provider keys
```
philcluff: Same

src/lib/providers.ts
```diff
@@ -0,0 +1,109 @@
...
+      requireEnv(apiKey, 'OPENAI_API_KEY');
+      const openai = createOpenAI({
+        apiKey,
+      });
+
+      return {
+        provider,
+        modelId,
+        model: openai(modelId),
+        usesMuxTokens: false,
```
philcluff: As above, remove mentions of potential future direction

src/functions/moderation.ts
```diff
@@ -0,0 +1,410 @@
...
+  const targetUrls =
+    submissionMode === 'base64'
+      ? (await downloadImagesAsBase64(imageUrls, downloadOptions, maxConcurrent)).map(
+          (img) => ({ url: img.url, image: img.base64Data })
+        )
+      : imageUrls.map((url) => ({ url, image: url }));
+
+  const moderate = async (entry: { url: string; image: string }): Promise<ThumbnailModerationScore> => {
+    try {
+      const res = await fetch('https://api.openai.com/v1/moderations', {
```
philcluff, line 154: I wasn't expecting to see direct calls to openai in here - is this something that's not abstracted by the AI sdk?

src/functions/moderation.ts
```diff
@@ -0,0 +1,410 @@
...
+  downloadOptions?: ImageDownloadOptions
+): Promise<ThumbnailModerationScore[]> {
+  const targetUrls =
+    submissionMode === 'base64'
+      ? (await downloadImagesAsBase64(imageUrls, downloadOptions, maxConcurrent)).map(
+          (img) => ({ url: img.url, image: img.base64Data })
+        )
+      : imageUrls.map((url) => ({ url, image: url }));
+
+  const prompt = `You are a content safety system. Analyze the provided image and return JSON with two numeric scores between 0 and 1: "sexual" and "violence". Higher = more risky.`;
```
philcluff: I am transparently reluctant to push customers in this direction honestly - letting a customer use a generic LLM for moderation feels wrong when our other vendors (open AI and hive) are explicit trained models for this workload.

src/functions/moderation.ts
```diff
@@ -0,0 +1,410 @@
...
+        provider as SupportedProvider
+      )
+    : null;
+
+  const muxClient = workflowClients?.mux || createMuxClient(validateCredentials(options));
+
+  // Fetch asset data from Mux
+  const asset = await muxClient.video.assets.retrieve(assetId);
+
+  // Get playback ID - prefer public playback IDs
```
philcluff: Move this logic into the playback ID helper, as we'll need that for signed URLs in the future.

src/primitives/storyboards.ts
```diff
@@ -0,0 +1,6 @@
+export const DEFAULT_STORYBOARD_WIDTH = 640;
+
+export function getStoryboardUrl(playbackId: string, width: number = DEFAULT_STORYBOARD_WIDTH): string {
+  return `https://image.mux.com/${playbackId}/storyboard.png?width=${width}`;
```
philcluff, line 4: Good opportunity for us to merge in private assets in the future here.

src/primitives/thumbnails.ts
```diff
@@ -0,0 +1,33 @@
...
+      timestamps.push(Math.round(i * spacing));
+    }
+  } else {
+    for (let time = 0; time < duration; time += interval) {
+      timestamps.push(time);
+    }
+  }
+
+  return timestamps.map(
+    (time) => `https://image.mux.com/${playbackId}/thumbnail.png?time=${time}&width=${width}`
```
philcluff, line 31: Same comment - good future opportunity for private playback ID abstraction

src/types.ts
```diff
@@ -1,13 +1,291 @@
...
   openaiApiKey?: string;
+  /** Anthropic API key (defaults to process.env.ANTHROPIC_API_KEY). */
   anthropicApiKey?: string;
+  /** Google Generative AI API key (defaults to GOOGLE_GENERATIVE_AI_API_KEY/GOOGLE_API_KEY). */
+  googleApiKey?: string;
+  /** Hive Visual Moderation API key (defaults to process.env.HIVE_API_KEY). */
+  hiveApiKey?: string;
+  /**
+   * Reserved for future hosted deployments that may require overriding the API
+   * base URL.
```
philcluff: Base URL of what?

src/types.ts
philcluff: Yeh, I think I agree with Rahim here

philcluff (changes requested):
Mostly LGTM, but there's few hard changes I'd like to see before we merge this

philcluff (changes requested):
Mostly LGTM, but there's few hard changes I'd like to see before we merge this

examples/burned-in-captions/provider-comparison.ts
cjpillsbury: Upvote.

src/types.ts
monsieurBoutte: Right on, I can make that change

examples/burned-in-captions/provider-comparison.ts
monsieurBoutte: yeah good point

examples/moderation/google-example.ts
monsieurBoutte: yeah I was thinking the same last night. Will do 👍

src/types.ts
monsieurBoutte: yeah I think I'm going to remove that for the time being. We can revisit how this manifest within the call-site, once we start experimenting with a different entry path like _"just use my mux ai credits"_ where we have more opinionated integrations with providers we've partnered with baked into some of these functions/workflows

src/functions/summarization.ts
monsieurBoutte: I'm going to hold off on making this change but certainly marking it as a follow up task for another PR to further flesh out

examples/moderation/hive-example.ts
philcluff: @monsieurBoutte Mind doing this one?

examples/moderation/README.md
```diff
@@ -1,6 +1,6 @@
 # Moderation Examples
 
-This directory contains examples demonstrating how to use the `getModerationScores` function from `@mux/ai`.
+This directory contains examples demonstrating how to use the `getModerationScores` helper from `@mux/ai/primitives`.
```
philcluff: Moderation is back in functions now, right?

monsieurBoutte: yup, let me fix that doc

src/functions/moderation.ts
philcluff: @monsieurBoutte mind replying to this one?

src/functions/moderation.ts
philcluff: @monsieurBoutte mind replying to this one?

src/functions/summarization.ts
philcluff: @monsieurBoutte mind replying to this one?

src/functions/translate-captions.ts
philcluff: @monsieurBoutte Mind doing this one?

src/lib/client-factory.ts
philcluff: @monsieurBoutte Mind replying to this one please?

src/functions/translate-audio.ts
monsieurBoutte: for now ¯_(ツ)_/¯

src/functions/summarization.ts
monsieurBoutte: good idea, i can add that detail

src/functions/summarization.ts
monsieurBoutte: yeah I agree, we should fail this request if it doesn't come back with a non-empty suggested string for the title

src/lib/client-factory.ts
monsieurBoutte: `GOOGLE_GENERATIVE_AI_API_KEY` is the more common one. I'll tweak it to look for that instead of both

src/lib/mux-assets.ts
monsieurBoutte: let me know if you have a suggested alternative or if something like `getPlaybackIdForAsset` sounds better

src/functions/moderation.ts
monsieurBoutte: ah yeah that's intentional because vercel's ai-sdk doesn't support that at the moment. So we'd otherwise have the entire openai SDK just for that moderation use case. Figured hitting their API was desirable and closer to what we do with Hive for the same reason

src/functions/moderation.ts
monsieurBoutte: yeah, I don't have any strong objections there. We can dial that back to only those two provider options

src/functions/translate-audio.ts
mihar-22: This can be a follow-up! 

Hard loops can sometimes behave unexpectedly in serverless functions, if it times out then the process will be killed  and we can't log an error. We might prefer using [p-retry](https://github.com/sindresorhus/p-retry) which is already installed. 

- It supports abort signal so we can stop retrying. Our current logic is not checking the abort so it means $$.
- We should set a `maxTimeout` that is less than the expected serverless function timeout. I think it's 10 seconds. This should be configurable for those that are using good ol' servers. The current 3mins is prob too long.

src/functions/index.ts
```diff
@@ -0,0 +1,6 @@
+export * from './summarization';
+export * from './translate-captions';
```
philcluff, line 2: Let's rename these two translation files to be consistent, I'd suggest:

`translate-captions`
`translate-audio`

src/functions/moderation.ts
philcluff: Sounds good!

README.md
philcluff: Yep, this one is on the backlog before release for sure.

philcluff (approved):
A few very minor follow ups, but this is overwhelmingly ready to merge!

src/functions/index.ts
philcluff: Pushed this change to the branch - what's one more change in this monster, eh?

src/functions/summarization.ts
monsieurBoutte: [opened an issue here](https://github.com/muxinc/ai/issues/13) to track this

src/functions/summarization.ts
monsieurBoutte: [issue created here](https://github.com/muxinc/ai/issues/18) to track progress on this

src/functions/summarization.ts
monsieurBoutte: [issue created here](https://github.com/muxinc/ai/issues/19) to track progress on this

src/functions/translate-audio.ts
monsieurBoutte: [issue created here](https://github.com/muxinc/ai/issues/20) for tracking progress

src/functions/translation.ts
monsieurBoutte: [issue created here](https://github.com/muxinc/ai/issues/21) to track progress on this

PR #6, opened 2025-11-24
test: Basic integration tests

philcluff:
Adds a basic set of integration tests, testing each of the main features.

Mostly oriented in happy path testing for now, does not pretend to have full coverage of functionality.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,40 @@
+name: Integration Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Run integration tests
+        env:
+          MUX_TOKEN_ID: ${{ secrets.MUX_TOKEN_ID }}
+          MUX_TOKEN_SECRET: ${{ secrets.MUX_TOKEN_SECRET }}
+          OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
+          ANTHROPIC_API_KEY: ${{ secrets.ANTHROPIC_API_KEY }}
+          ELEVENLABS_API_KEY: ${{ secrets.ELEVENLABS_API_KEY }}
+          HIVE_API_KEY: ${{ secrets.HIVE_API_KEY }}
+          S3_ENDPOINT: ${{ secrets.S3_ENDPOINT }}
+          S3_REGION: ${{ secrets.S3_REGION }}
+          S3_BUCKET: ${{ secrets.S3_BUCKET }}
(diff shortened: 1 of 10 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+jobs:
+  integration-tests:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
monsieurBoutte, line 12: 🔥

tests/integration/burned-in-captions.test.ts
```diff
@@ -0,0 +1,101 @@
...
+describe('Burned-in Captions Integration Tests', () => {
+  const assetsWithCaptions = [
+    'atuutlT45YbyucKU15u0100p45fG2CoXfJOd02VWMg4m004',
+    'gEvCHSJRioaSMHtsJxT4DA02ee3xbgVL02sDGZJuqt01vs',
+  ];
+
+  const assetsWithoutCaptions = [
+    'gIRjPqMSRcdk200kIKvsUo2K4JQr6UjNg7qKZc02egCcM',
+    // '88Lb01qNUqFJrOFMITk00Ck201F00Qmcbpc5qgopNV4fCOk', // FIXME: This asset incorrectly detects captions; needs investigation and prompt tuning
+  ];
```
monsieurBoutte, line 14: 🔧 Minor concern here about the value prop of hardcoded asset ids in this way. I wonder if we should lift them into env variables that can be injected locally by devs and automatically in ci/cd

tests/integration/burned-in-captions.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      const result = await hasBurnedInCaptions(assetId, {
+        provider: 'openai',
+      });
+
+      // Assert that the result exists
+      expect(result).toBeDefined();
+      expect(result.hasBurnedInCaptions).toBe(true);
+
+      // Assert confidence is greater than 80%
+      expect(result.confidence).toBeGreaterThan(0.8);
```
monsieurBoutte, line 27: 💭 Minor nit here: this is fine to leave for now, but I'd advocate for moving these types of assertions into evaluations. Asserting the confidence of a workflow fits more appropriately in the evaluation territory rather than in integration tests.

tests/integration/chapters.test.ts
```diff
@@ -0,0 +1,56 @@
+import { describe, it, expect } from 'vitest';
+import 'dotenv/config';
+import { generateChapters } from '../../src/chapters';
+
+describe('Chapters Integration Tests', () => {
+  const assetId = '88Lb01qNUqFJrOFMITk00Ck201F00Qmcbpc5qgopNV4fCOk';
```
monsieurBoutte, line 6: 💭 similar note here as i left above about hardcoded asset ids

monsieurBoutte (approved):
nice work

tests/integration/burned-in-captions.test.ts
philcluff: Yeh, totally agree, we can rip this specific assert once we have evals!

tests/integration/burned-in-captions.test.ts
philcluff: Yeh - I think I'm going to leave them for now - I'd maybe propose a `fixtures.yaml/json` file instead to store these, rather than env vars. We'll have to tackle this in evals too.

PR #32, opened 2025-12-02
chore: Re-lint repo now we have commit checks

philcluff:


src/lib/prompt-builder.ts
```diff
@@ -69,38 +69,38 @@ export interface PromptBuilder<TSections extends string> {
 export function renderSection(section: PromptSection): string {
   const { tag, content, attributes } = section;
 
-  const XML_NAME_PATTERN = /^[A-Za-z_][A-Za-z0-9_.:-]*$/;
```
monsieurBoutte, line 72: Did we verify the test still pass here?

monsieurBoutte (approved):
Looks good but left one comment about the regex change

src/lib/prompt-builder.ts
philcluff: They doooo, but I must admit I'm a bit surprised by this change coming from the linter.

philcluff: For audit: 

We do use eslint-plugin-regexp rules that come with @antfu/eslint-config. Specifically, these rules caused the change:

  Enabled rules (all set to error level 2):

  1. regexp/use-ignore-case - Suggests using the /i flag instead of [A-Za-z]
    - Changed: [A-Za-z_] → [A-Z_] with /i flag
  2. regexp/prefer-w - Suggests using \w instead of [A-Za-z0-9_]
    - Changed: [A-Za-z0-9_] → \w
  3. regexp/prefer-d - Suggests using \d instead of [0-9]
    - (Not applied in your case, but available)

These rules are part of @antfu/eslint-config's default setup and help optimize regex patterns for readability and performance.

Issue #14, opened 2025-11-26
Add evals for all AI workflows


monsieurBoutte:
# Add evals for most AI workflows

---

## Description

Create an evaluation framework to measure output quality and consistency across our AI-powered workflows.

## Background

We currently have integration tests for functional correctness, but lack evaluations (evals) that measure the **quality** of AI outputs. As we iterate on prompts, switch models, or add providers, we need a systematic way to detect regressions and compare performance.

## Workflows to evaluate

| Workflow | Output to evaluate |
|----------|-------------------|
| `summarization` | Title, description, and tag relevance/accuracy |
| `moderation` | Sexual/violence score precision vs ground truth |
| `chapters` | Chapter boundary accuracy and title quality |
| `burned-in-captions` | Caption rendering correctness |
| `translate-captions` | Translation accuracy and fluency |
| `translate-audio` | TBD |

## Acceptance Criteria

- [ ] Eval dataset fixtures with labeled examples (input assets + expected/rated outputs)
- [ ] Scoring functions per workflow type (e.g., similarity, precision, human-preference proxy)
- [ ] Support for comparing results across providers (OpenAI, Anthropic, etc.)
- [ ] CI-integrated eval runner with configurable pass/fail thresholds
- [ ] Eval results output (JSON report, markdown summary, or dashboard)

## Suggested Approach

1. Start with `summarization` and `moderation` as pilot workflows
2. Use a small curated dataset (~3-10 examples per workflow)
3. Define simple metrics first (exact match, factuality, threshold accuracy)
4. Expand to LLM-as-judge for subjective quality

## Related

- Existing integration tests: `tests/integration/`
- Function implementations: `src/functions/`

PR #39, opened 2025-12-03
chore: initial summarization evals

monsieurBoutte:
## Summary

Added evaluation coverage for the `getSummaryAndTags` workflow to measure quality, performance, and cost across AI providers. Which helps address some of #14 as well

### 1. New Eval Coverage (`tests/eval/summarization.eval.ts`)

Comprehensive eval using [evalite](https://github.com/mattpocock/evalite) that tests summarization across OpenAI, Anthropic, and Google providers.

**Metrics tested:**

| Category   | Scorers                                                                                                                           |
| ---------- | --------------------------------------------------------------------------------------------------------------------------------- |
| Efficacy   | title-quality, title-similarity, description-quality, tags-quality, response-integrity, no-filler-phrases, description-similarity |
| Efficiency | latency-performance (<5s good), token-efficiency (<4k tokens)                                                                     |
| Expense    | usage-data-present, cost-within-budget (<$0.005/request)                                                                          |

Uses semantic similarity (embeddings) for flexible title/description comparison against reference answers.

### 2. Workflow Enhancements (`src/workflows/summarization.ts`)

- **Added token usage tracking**: `SummaryAndTagsResult` now includes `usage?: TokenUsage`
- **Added transcript passthrough**: Returns `transcriptText` when transcript was used
- **Improved prompt**: Added `<language_guidelines>` to prevent meta-descriptive phrases ("The image shows...", "In this video...")

### 3. Simplified Integration Tests (`tests/integration/summarization.test.ts`)

Refactored from 3 duplicate tests to single parameterized test using `it.each(providers)`. Detailed assertions moved to eval suite.

## Running the Evals

```bash
npx evalite serve summarization.eval.ts
```

Results viewable at `http://localhost:3006` after running.

### DIFF
--- a/src/workflows/summarization.ts
+++ b/src/workflows/summarization.ts
@@ -19,7 +19,7 @@ import { withRetry } from "../lib/retry";
 import { resolveSigningContext } from "../lib/url-signing";
 import { getStoryboardUrl } from "../primitives/storyboards";
 import { fetchTranscriptForAsset } from "../primitives/transcripts";
-import type { ImageSubmissionMode, MuxAIOptions, ToneType } from "../types";
+import type { ImageSubmissionMode, MuxAIOptions, TokenUsage, ToneType } from "../types";
 
 // ─────────────────────────────────────────────────────────────────────────────
 // Types
@@ -47,6 +47,10 @@ export interface SummaryAndTagsResult {
   tags: string[];
   /** Storyboard image URL that was analyzed. */
   storyboardUrl: string;
+  /** Token usage from the AI provider (for efficiency/cost analysis). */
+  usage?: TokenUsage;
+  /** Raw transcript text used for analysis (when includeTranscript is true). */
+  transcriptText?: string;
 }
 
 /**
@@ -191,7 +195,22 @@ const SYSTEM_PROMPT = dedent`
     - Only describe what is clearly observable in the frames or explicitly stated in the transcript
     - Do not fabricate details or make unsupported assumptions
     - Return structured data matching the requested schema
-  </constraints>`;
+  </constraints>
+
+  <language_guidelines>
+    AVOID these meta-descriptive phrases that reference the medium rather than the content:
+    - "The image shows..." / "The storyboard shows..."
+    - "In this video..." / "This video features..."
+    - "The frames depict..." / "The footage shows..."
+    - "We can see..." / "You can see..."
+    - "The clip shows..." / "The scene shows..."
+
+    INSTEAD, describe the content directly:
+    - BAD: "The video shows a chef preparing a meal"
+    - GOOD: "A chef prepares a meal in a professional kitchen"
+
+    Write as if describing reality, not describing a recording of reality.
+  </language_guidelines>`;
 
 interface UserPromptContext {
   tone: ToneType;
@@ -307,7 +326,12 @@ export async function getSummaryAndTags(
   // Analyze storyboard with AI provider (signed if needed)
   const imageUrl = await getStoryboardUrl(playbackId, 640, policy === "signed" ? signingContext : undefined);
 
-  const analyzeStoryboard = async (imageDataUrl: string) => {
+  interface AnalysisResponse {
+    result: SummaryType;
+    usage: TokenUsage;
+  }
+
+  const analyzeStoryboard = async (imageDataUrl: string): Promise<AnalysisResponse> => {
     const response = await generateObject({
       model: clients.languageModel.model,
       schema: summarySchema,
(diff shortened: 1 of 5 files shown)


tests/eval/burned-in-captions.eval.ts
```diff
@@ -278,10 +289,10 @@ evalite("Burned-in Captions", {
 
     // LATENCY: Wall clock time performance
     // Measures: Time from request initiation to response completion
-    // Score: 1.0 for <7s, linear decline to 0.5 at 15s, continues declining toward 0 beyond 15s
+    // Score: 1.0 for <5s, linear decline to 0.5 at 12s, continues declining toward 0 beyond 12s
     {
       name: "latency-performance",
-      description: `Scores latency: 1.0 for <${LATENCY_THRESHOLD_GOOD_MS}ms, scaled down to 0 for >${LATENCY_THRESHOLD_ACCEPTABLE_MS}ms.`,
+      description: `Scores latency: 1.0 for <5000ms, scaled down to 0 for >12000ms.`,
```
cjpillsbury, line 295: nitpick: Why remove parameterization from here?

tests/eval/summarization.eval.ts
```diff
@@ -0,0 +1,546 @@
...
+ * 4. RESPONSE INTEGRITY
+ *    - All required fields populated correctly
+ *    - Types match expected schema
+ *    - Storyboard URL properly generated
+ *
+ * 5. NO FILLER PHRASES
+ *    - Description should NOT contain "the image shows", "the video shows", etc.
+ *    - Content should be described directly without meta-references to the medium
+ *
+ * 6. TITLE SIMILARITY (embeddings)
```
cjpillsbury: praise: 👌 (and below use of embeddings)

tests/eval/summarization.eval.ts
```diff
@@ -0,0 +1,546 @@
...
+ *    - Reasonable length (typically under 10 words / 100 chars)
+ *    - Does NOT start with "A video of" or similar phrasing
+ *    - Uses active, specific language
+ *
+ * 2. DESCRIPTION QUALITY
+ *    - Non-empty, meaningful summary
+ *    - Reasonable length (2-4 sentences, up to 1000 chars)
+ *    - Describes observable content
+ *
+ * 3. TAGS QUALITY
```
cjpillsbury, line 37: idea(non-blocking): consider defining reference tags and using embeddings across the set for a more objective eval here, too.

tests/eval/burned-in-captions.eval.ts
monsieurBoutte: unintended move, good call out

tests/eval/summarization.eval.ts
```diff
@@ -0,0 +1,546 @@
...
+    const result = await getSummaryAndTags(assetId, {
+      provider,
+      includeTranscript: true,
+    });
+    const latencyMs = performance.now() - startTime;
+
+    const usage = result.usage ?? {};
+    const estimatedCostUsd = calculateCost(
+      provider,
+      usage.inputTokens ?? 0,
```
cjpillsbury, line 238: question(non-blocking): Does this mean we don't/can't distinguish between cases where we're missing token data vs. cases where there were no tokens used?

tests/eval/summarization.eval.ts
```diff
@@ -0,0 +1,546 @@
...
+        totalTokens: usage.totalTokens ?? 0,
+      },
+      start: startTime,
+      end: startTime + latencyMs,
+    });
+
+    return {
+      ...result,
+      provider,
+      model: DEFAULT_LANGUAGE_MODELS[provider],
```
cjpillsbury, line 258: question(non-blocking): Confused by this. If we're using evals to in part determine what the default models should be, shouldn't this be parameterized and not explicitly referencing DEFAULT_LANGUAGE_MODELS?

monsieurBoutte: no so what this is doing is ensuring that all these evals trace back to the current `SupportedProvider` and their respective default models.

tests/eval/summarization.eval.ts
```diff
@@ -0,0 +1,546 @@
...
+          return 0;
+        }
+
+        // Must be reasonable length
+        if (title.length > TITLE_MAX_LENGTH) {
+          return 0.5;
+        }
+
+        // Should NOT start with filler phrases
+        const fillerPhrases = [
```
cjpillsbury, line 313: idea(non-blocking): doing programmatic exact matches is always a bit brittle wrt false negatives. This seems reasonable as a first pass and probably is "good enough", but we might be able to get clever with this using text embedding distances on reference phrase(s) as well. Lower priority, but also there may actually be content that is _**about**_ "a video" that we want to eval, in which case, this would be a false positive match.

tests/eval/summarization.eval.ts
monsieurBoutte: this helps with more informed decisions about what's the specified default model at the workflow level based on the benchmarks ran within the corresponding eval suite for said workflow

src/workflows/summarization.ts
```diff
@@ -327,18 +351,27 @@ export async function getSummaryAndTags(
       ],
     });
 
-    return response.object;
+    return {
+      result: response.object,
+      usage: {
```
cjpillsbury, line 356: idea(non-blocking): It feels wrong/unfortunate that we changed our response shape for what appears to be solely for evals. I'm wondering if we could instead provide an optional callback or similar as a second param for this kind of data?

tests/eval/summarization.eval.ts
monsieurBoutte: no this is just to make typescript happy.

cjpillsbury (approved):
pre-approving with only one nitpick and some general (non-blocking) ideas/questions.

tests/eval/summarization.eval.ts
cjpillsbury: kk. just had some worries about bugs resulting in actual undefined tokens then showing up as e.g. 0$ in programmatic evals, resulting in inflated scores (worse, inflated scores potentially tied to issues with a provider)

src/workflows/summarization.ts
monsieurBoutte: I think including this is very useful in the response. Not just for evals but for general expense attributions that our consumers are likely on the hook for tracking

tests/eval/summarization.eval.ts
monsieurBoutte: I'll add coverage to ensure we're getting greater than zero token counts here, especially if cachedTokens is also zero

PR #42, opened 2025-12-04
docs: Initial pass of restructuring docs

philcluff:


docs/API.md
```diff
@@ -0,0 +1,376 @@
...
+
+```typescript
+{
+  assetId: string;
+  hasBurnedInCaptions: boolean; // Whether burned-in captions were detected
+  confidence: number; // Confidence score (0.0-1.0)
+  detectedLanguage: string | null; // Language of detected captions, or null
+  storyboardUrl: string; // URL to analyzed storyboard
+}
+```
```
monsieurBoutte, line 138: 🔧 not blocking but this is missing the new usage telemetry we're now returning. I'm in the process of adding it to summaries and the plan is to had it to all areas we're integrating with vercels ai-sdk

docs/EXAMPLES.md
```diff
@@ -0,0 +1,212 @@
...
+**Translation Workflow:**
+
+1. Fetches existing captions from Mux asset
+2. Translates VTT content using your selected provider (default: Claude Sonnet 4.5)
+3. Uploads translated VTT to S3-compatible storage
+4. Generates presigned URL (1-hour expiry)
+5. Adds new subtitle track to Mux asset
+6. Track name: "{Language} (auto-translated)"
+
+## Audio Dubbing Examples
```
monsieurBoutte, line 162: for dubbing and audio translation, we might consider adding a note that you can also verify this on the player within the users mux account. It's one thing to check the assets exist but it's nice to encourage them to go see/listen to the output applied in the player. It's a pretty neat experience for the elevan labs workflow especially

docs/PRIMITIVES.md
```diff
@@ -0,0 +1,518 @@
+# Primitives
+
+Low-level building blocks for constructing custom AI workflows with Mux video data.
+
+## Overview
+
+Primitives are focused utilities that fetch and transform Mux video data without making AI provider calls. Use these when you need to:
```
monsieurBoutte, line 7: 💭 idk that "without making AI provider calls" will always be true but probably not worth thinking too deeply about at present

docs/WORKFLOWS.md
```diff
@@ -0,0 +1,401 @@
+# Workflows
+
+Detailed documentation for each pre-built workflow.
```
monsieurBoutte, line 3: 🔧 I think we're missing an opportunity here to include a little more prose that spell out this our orchestration layer that composes over our primitives. Perhaps a little redundant give docs in other areas but necessary in my opinion, given that this is an isolated file

monsieurBoutte (approved):
Looks good for the most part. A couple minor nits. Similar to the comment I left for the workflows markdown file, I see an opportunity at the top of the README to clarify the value proposition for incoming readers. We could add more detail about the currently exposed layers, such as mux primitives and the workflows that build upon those primitives.

README.md
monsieurBoutte, line 25: we can make this update after changing the default models but given the inconsistency in of Haiku's structured output — I think we should say something like `Quality by default`

docs/EXAMPLES.md
philcluff: Good shout, done!

docs/PRIMITIVES.md
philcluff: Agreed - we can strip it out if it changes!

docs/API.md
philcluff: 👍🏻  Sounds good, let's add in a follow up

docs/WORKFLOWS.md
philcluff: Done!

PR #48, opened 2025-12-04
fix: patch for downloading translated audio

monsieurBoutte:
So it appears that Eleven Labs supports both 2-letter codes and 3-letter codes. But [in their documentation, they call out support only for 3-letter codes](https://elevenlabs.io/docs/capabilities/speech-to-text#supported-languages). And confusingly, when you go to download translated audio from their API, you have to use a 2-letter code.


cjpillsbury (approved):
👍

PR #49, opened 2025-12-05
chore: publish evals as main changes

monsieurBoutte:
## Summary

This PR adds **automated eval publishing infrastructure** so that every push to `main` runs our AI workflow evaluations and publishes the results to GitHub Pages.

### Changes

1. **New GitHub Actions Workflow** (`.github/workflows/publish-evals-ui.yml`)
   - Runs `npx evalite run` on every push to `main`
   - Exports the Evalite UI as static files
   - Deploys to GitHub Pages at https://muxinc.github.io/ai/evals

2. **New Evals Documentation** (`docs/EVALS.md`)
   - Documents our "3 E's" eval framework: **Efficacy**, **Efficiency**, **Expense**
   - Explains how to run evals locally (`npm run test:eval`)
   - Provides guidance on adding evals for new workflows
   - Includes model pricing reference table

3. **Test Timeout Increase** (`vitest.config.ts`)
   - Bumped from 60s → 120s to accommodate longer-running integration tests. There were a few flakes pop up recently while we still had a 60s timeout.

## Why

- **Visibility**: Persistent dashboard of eval results at a public URL
- **Standards**: Establishes that no new workflow ships without eval coverage
- **Decision-making**: Enables data-driven model selection across OpenAI, Anthropic, and Google


docs/EVALS.md
```diff
@@ -0,0 +1,224 @@
+# Evaluations
```
philcluff, line 1: Note: Will need to merge this docs into @cjpillsbury's docs PR

.github/workflows/publish-evals-ui.yml
```diff
@@ -0,0 +1,60 @@
...
+concurrency:
+  group: pages
+  cancel-in-progress: true
+
+jobs:
+  build-and-deploy:
+    runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
```
philcluff, line 22: Any concerns with this info being public, or risk of information leak?

philcluff (approved):

.github/workflows/publish-evals-ui.yml
monsieurBoutte: I don't think so. It's a static HTML page that gets deployed after all the evals have ran. The secrets are safe. But we may want to give some thought around whether or not we want asset ids exposed in the eval results.

PR #52, opened 2025-12-09
Make workflows durable

dylanjha:
What this PR aims to do:

1. Update the top level `workflows/` exports to be compatible with [Workflow DevKit](https://useworkflow.dev/)
2. Add a new set of integration tests (still through vitest, with a different vitest config that runs all the same `workflows/` functions, but backed by Workflow DevKit (queuing, running steps, etc. all happen)

In order to achieve (1) we have to:

- Make sure all node dependencies are only done in `"use step"` functions. [See here](https://useworkflow.dev/docs/foundations/workflows-and-steps#workflow-functions)
  - > Workflow functions define the entrypoint of a workflow and organize how step functions are called. This type of function does not have access to the Node.js runtime, and usable npm packages are limited.
- Add `"use workflow"` directive to top level workflow functions
- Add `"use step"` directive to async functions within workflows
- Make all `"use step"` functions serializable. [See here](https://useworkflow.dev/docs/foundations/serialization)
  - > All function arguments and return values passed between workflow and step functions must be serializable. Workflow DevKit uses a custom serialization system built on top of [devalue](https://github.com/sveltejs/devalue). This system supports standard JSON types, as well as a few additional popular Web API types.
  - You CANNOT pass around references to functions or instances of classes, etc.
  - ❗ This is the most significant refactor

In "normal" node env, these directives are a no-op

In a Workflow DevKit world, these functions make the workflow durable, and are backed up with a nice UI for observability.

In order to achieve (2) we have to:

- Configure [Vitest kinda like the Vite guide](https://useworkflow.dev/docs/getting-started/vite)
- Run a `nitro` webserver and sidecar process to actually run the queue

# TODO

- [x] add back env loading to handle `.env.test`
- [x] remove debug logs
- [x] debug why our regular tests aren't passing
- [x] lint / format / fix
- [x] add back the translation workflows that depend on aws-sdk package, add that package back
- [x] add back the `DEFAULT_PROVIDER` constant -- not sure why that was giving me trouble in a build
- [x] add the rest of the workflows integration tests
- [x] add workflows integration tests to CI
- [x] code review for architecture, structure, repetition, etc

# Other notes:

- [x] bump node version in CI to 22 (blocker for one of the dev dependencies)


src/lib/image-download.ts
```diff
@@ -61,13 +61,11 @@ export async function downloadImageAsBase64(
   url: string,
   options: ImageDownloadOptions = {},
 ): Promise<ImageDownloadResult> {
+  "use step";
   const opts = { ...DEFAULT_OPTIONS, ...options };
-  let attemptCount = 0;
-
   return pRetry(
     async () => {
-      attemptCount++;
```
dylanjha, line 69: TODO -- add back this attemptCount functionality

src/lib/mux-assets.ts
```diff
@@ -29,9 +31,15 @@ function getPlaybackId(asset: MuxAsset): { id: string; policy: PlaybackPolicy }
...
 export async function getPlaybackIdForAsset(
-  mux: Mux,
+  credentials: ValidatedCredentials,
   assetId: string,
 ): Promise<PlaybackAsset> {
+  "use step";
+  const mux = new Mux({
+    tokenId: credentials.muxTokenId,
+    tokenSecret: credentials.muxTokenSecret,
+  });
```
dylanjha, line 41: NOTE FOR REVIEWERS:

You'll see this pattern a lot -- instead of passing in a `mux` instance (a non-primitive), we pass in `credentials` (a primitive, play JS object)

src/workflows/burned-in-captions.ts
```diff
@@ -178,14 +180,13 @@ function buildUserPrompt(promptOverrides?: BurnedInCaptionsPromptOverrides): str
 // Implementation
 // ─────────────────────────────────────────────────────────────────────────────
 
-const DEFAULT_PROVIDER = "openai";
```
dylanjha, line 181: TODO -- I'm not sure why, but something with the bundler was breaking with this `const DEFAULT_PROVIDER` at the top level, I have to revisit it.

src/workflows/chapters.ts
```diff
@@ -116,18 +117,28 @@ export async function generateChapters(
...
       generateObject({
-        model: clients.languageModel.model,
+        model,
         schema: chaptersSchema,
         abortSignal,
         messages: [
           {
             role: "system",
-            content: SYSTEM_PROMPT,
+            content: systemPrompt,
```
dylanjha: TODO -- review `SYSTEM_PROMPT` constant here and the `systemPrompt` var -- not sure why this got changed

src/lib/providers.ts
```diff
@@ -2,7 +2,7 @@ import { createAnthropic } from "@ai-sdk/anthropic";
 import { createGoogleGenerativeAI } from "@ai-sdk/google";
 import { createOpenAI } from "@ai-sdk/openai";
 
-import env from "../env";
+import env from "../env.ts";
```
dylanjha, line 5: I hate this change, but I can't find another way to make this work & make nitro happy in CI

src/workflows/embeddings.ts
```diff
@@ -38,14 +39,6 @@ export interface EmbeddingsOptions extends MuxAIOptions {
 // Implementation
 // ─────────────────────────────────────────────────────────────────────────────
 
-const DEFAULT_PROVIDER = "openai";
-const DEFAULT_CHUNKING_STRATEGY: ChunkingStrategy = {
-  type: "token",
-  maxTokens: 500,
-  overlap: 100,
-};
-const DEFAULT_BATCH_SIZE = 5;
```
dylanjha, line 47: this change is necessary because functions with `"use workflow"` need to have all their variables inside the scope of the function

src/env.ts
```diff
@@ -1,16 +1,17 @@
...
-import { expand } from "dotenv-expand";
 import { z } from "zod";
 
-expand(config({
-  path: path.resolve(
-    process.cwd(),
-    process.env.NODE_ENV === "test" ? ".env.test" : ".env",
-  ),
-}));
+// import { config } from "dotenv";
```
cjpillsbury: nitpick(non-blocking): remove commented out code (we can use commit history if we need to recall this).

src/lib/providers.ts
cjpillsbury: There's an optional followup here that we shouldn't treat as blocking that @monsieurBoutte was working on

test-server/nitro.config.ts
```diff
@@ -0,0 +1,8 @@
+import { defineNitroConfig } from "nitro/config";
+
+export default defineNitroConfig({
```
cjpillsbury, line 8: thought(non-blocking): through code comments or elsewhere, let's make sure we document our woes and why this (and related) _**had to be**_ moved into a sub-directory a la `test-server`

cjpillsbury (approved):
We can merge the creds crud directly to main as a followup. Only non-blocking feedback! Feel free to respond and self-merge as you see fit.

src/workflows/translate-captions.ts
```diff
@@ -77,28 +77,177 @@ export type TranslationPayload = z.infer<typeof translationSchema>;
...
+  return trackResponse.id;
+}
 
 export async function translateCaptions<P extends SupportedProvider = SupportedProvider>(
   assetId: string,
   fromLanguageCode: string,
   toLanguageCode: string,
   options: TranslationOptions<P>,
 ): Promise<TranslationResult> {
+  "use workflow";
```
monsieurBoutte, line 239: 💭 through the use of code comments across these files, i think it would be helpful to clearly call out where steps end and workflows begin

monsieurBoutte (approved):
amazing work on this

PR #50, opened 2025-12-05
docs: update readme and create basic contributing docs

cjpillsbury:
This is primarily docs updates.

# `README.md`
1. Updated the readme to front load information for folks new to the package and potentially new to mux.
2. Cleaned up some unclear parts in the prior docs version
3. Added some additional info that wasn't here before
4. Split out some of the documentation into a separate CONTRIBUTING.md

**NOTE:** Per a convo with @philcluff, I kept links to content in this repo even though it is currently private, since we plan to make it public before GA.

# `CONTRIBUTING.md`
1. Wrote a reasonable starting point for `CONTRIBUTING.md`

# Other
1. Updated some env crud (including GHA) for consistency and to target node >= 21 based on fetch API stabilization
2. Added a reference `.env.example`

# Out of scope good fast follows
1. Test/validate/document usage with: Deno, bun
2. Additional passes on docs for API and related.

PR #56, opened 2025-12-10
Christian & dylan hacking sesh

dylanjha:
- [x]  full suite of integraiton tests for every workflow in a Workflow DevKit environment (I basically duplicated every workflow integration test and called it with `start(workflowFunction, [args..])`
- [x]  there are now 2 integraiton test commands `npm run test:integration` and `npm run test:integration-workflowdevkit` -- each one requires a different configuration of vitest
- [x] updated README.md
- [x] add DEVELOPMENT.md


vitest.workflowdevkit.setup.ts
```diff
@@ -53,9 +53,11 @@ export async function setup() {
   // eslint-disable-next-line no-console
   console.log("Nitro server started and ready for workflow execution");
 
-  // Set the base URL for local workflow execution
+  // Set the base URL and data dir for local workflow execution
   // eslint-disable-next-line node/no-process-env
   process.env.WORKFLOW_LOCAL_BASE_URL = `http://localhost:${PORT}`;
+  // eslint-disable-next-line node/no-process-env
+  process.env.WORKFLOW_LOCAL_DATA_DIR = "./test_server/.workflow-data";
```
dylanjha: this is key, so that workflow data is written to the correct place

vitest.workflowdevkit.setup.ts
```diff
@@ -15,7 +15,7 @@ export async function setup() {
   nitroServer = spawn("npx", ["nitro", "dev", "--port", PORT], {
     stdio: "pipe",
     detached: false,
-    cwd: process.cwd(),
+    cwd: "test_server",
```
dylanjha: this is key, so that the nitro server isn't slurping up all the project files and processing them

src/lib/providers.ts
```diff
@@ -2,7 +2,7 @@ import { createAnthropic } from "@ai-sdk/anthropic";
 import { createGoogleGenerativeAI } from "@ai-sdk/google";
 import { createOpenAI } from "@ai-sdk/openai";
 
-import env from "../env";
+import env from "../env.ts";
```
dylanjha, line 5: still need to debug why this is necessary

DEVELOPMENT.md
```diff
@@ -0,0 +1,42 @@
+# Developing @mux/ai 📼 🤝 🤖
```
cjpillsbury: issue(blocking): We've got a CONTRIBUTING.md. This feels like it is redundant. Let's add the workflow devkit-specific info there, but let's also try to make sure we're intentional about those updates.

cjpillsbury: Let's treat docs updates as a separate PR to make this a bit easier.

README.md
```diff
@@ -32,6 +32,80 @@ A typescript library for connecting videos in your Mux account to multi-modal LL
 - **Composable Building Blocks**: Import primitives to fetch transcripts, thumbnails, and storyboards to build bespoke flows
 - **Universal Language Support**: Automatic language name detection using `Intl.DisplayNames` for all ISO 639-1 codes
 
+## Automatically Compatible with Workflow DevKit / `"use workflow"`
```
cjpillsbury: issue(blocking): @philcluff and I were very intentional about the structure (order, terseness, clarity, etc.) of the README.md (also CONTRIBUTING.md, but lower level of concern there), including both out of band discussions and convo here https://github.com/muxinc/ai/pull/50. Let's treat docs updates as a separate PR, because I'm noticing that this is based off of out of date docs.

src/workflows/embeddings.ts
```diff
@@ -74,59 +67,38 @@ function averageEmbeddings(embeddings: number[][]): number[] {
...
  */
-async function generateChunkEmbeddings(
-  chunks: TextChunk[],
-  provider: SupportedEmbeddingProvider,
-  modelId: string,
-  credentials: ValidatedCredentials,
-  batchSize: number,
-  abortSignal?: AbortSignal,
-): Promise<ChunkEmbedding[]> {
+async function generateSingleChunkEmbedding({
```
cjpillsbury, line 72: nitpick(blocking): don't loose jsdocs.

src/workflows/chapters.ts
```diff
@@ -117,49 +158,16 @@ export async function generateChapters(
...
-      workflowConfig.provider,
-      workflowConfig.modelId,
-      workflowConfig.credentials,
-    );
-
-    return await withRetry(() =>
-      generateObject({
-        model,
-        schema: chaptersSchema,
-        abortSignal,
```
cjpillsbury, line 137: question(blocking): a bit worried we lost `abortSignal` here, since that is a functional change. would be good to get @monsieurBoutte's 👀 / 🧠  on this one.

src/workflows/chapters.ts
```diff
@@ -117,49 +158,16 @@ export async function generateChapters(
...
-    );
-  };
-
-  let chaptersData: { chapters: Chapter[] } | null = null;
+  let chaptersData: ChaptersType | null = null;
 
   try {
-    const response = await generateChaptersStep(
-      config,
+    chaptersData = await generateChaptersWithAI({
```
cjpillsbury, line 164: question(blocking): any reason for the rename here? If not, use original name.

src/workflows/embeddings.ts
```diff
@@ -169,9 +141,8 @@ export async function generateVideoEmbeddings(
     provider = "openai",
     model,
     languageCode,
-    chunkingStrategy = DEFAULT_CHUNKING_STRATEGY,
-    batchSize = DEFAULT_BATCH_SIZE,
-    abortSignal,
+    chunkingStrategy = { type: "token", maxTokens: 500, overlap: 100 } as ChunkingStrategy,
```
cjpillsbury: question(blocking): It seems like we're inconsistent with things being changed to/from default const values vs. inlines... any reason why? If these changes aren't necessary, can we either:
1. revert them to avoid unnecessary changes
2. refactor, but consistently use one approach or another

src/workflows/translate-audio.ts
```diff
@@ -135,7 +138,7 @@ async function waitForAudioStaticRendition({
   }
 
   for (let attempt = 1; attempt <= STATIC_RENDITION_MAX_ATTEMPTS; attempt++) {
-    await delay(STATIC_RENDITION_POLL_INTERVAL_MS);
+    await sleep(STATIC_RENDITION_POLL_INTERVAL_MS);
```
cjpillsbury, line 141: question(blocking): any reason for the rename here? If not, use original name.

src/workflows/translate-audio.ts
```diff
@@ -320,23 +470,16 @@ export async function translateAudio(
   let targetLanguages: string[] = [];
 
   while (dubbingStatus === "dubbing" && pollAttempts < maxPollAttempts) {
-    await new Promise(resolve => setTimeout(resolve, 10000)); // Wait 10 seconds
+    await sleep(10000); // Wait 10 seconds
```
cjpillsbury, line 473: praise: robot or otherwise, good catch of something that should have been using delay/sleep for consistency.
thought(non-blocking): As a follow up, since this is a perf concern, we should look to improve this (out of scope for this effort cc @monsieurBoutte).

src/workflows/translate-audio.ts
```diff
@@ -376,36 +519,29 @@ export async function translateAudio(
 
     // Find the matching language code from ElevenLabs response
     // First try exact match, then try case-insensitive match
-    let elevenLabsLangCode = targetLanguages.find(
+    let downloadLangCode = targetLanguages.find(
```
cjpillsbury, line 522: nitpick(blocking): let's revert this change to reduce noise as it is clearly unnecessary

src/workflows/translate-audio.ts
```diff
@@ -414,47 +550,19 @@ export async function translateAudio(
...
-
-    // Generate presigned URL (valid for 1 hour)
-    const getObjectCommand = new GetObjectCommand({
-      Bucket: s3Bucket!,
-      Key: audioKey,
-    });
-
-    presignedUrl = await getSignedUrl(s3Client, getObjectCommand, {
-      expiresIn: 3600, // 1 hour
+    presignedUrl = await uploadDubbedAudioToS3({
```
cjpillsbury, line 556: praise: 😎

tsconfig.json
```diff
@@ -13,6 +13,7 @@
       "@mux/ai/types": ["src/types.ts"]
     },
     "resolveJsonModule": true,
+    "allowImportingTsExtensions": true,
```
cjpillsbury, line 16: thought(non-blocking): assuming this is related to `env.ts` change we saw as necessary (without full understanding from either of us as to why). Just calling out for if/when we do followup investigation there.

vitest.workflowdevkit.setup.ts
cjpillsbury: nitpick(non-blocking): is there a naming convention for directory names we should stick to here (e.g. kebab vs snake)?

cjpillsbury (changes requested):
some minor change requests. overall 😎. aside from strong ask to do docs updates as a separate effort, any changes could be applied after this is merged to your base branch (as long as they're addressed before merging to #main)

src/lib/providers.ts
monsieurBoutte: probably just need to add ts-alias in the dev dependencies and then add to the tsconfig

```json
  "tsc-alias": {
    "resolveFullPaths": true
  }
  ```

src/workflows/translate-audio.ts
cjpillsbury: Ah just saw [sleep](https://useworkflow.dev/docs/api-reference/workflow/sleep). This is better even if it doesn't get us all the way. Resolving.

src/lib/providers.ts
monsieurBoutte: added the commit for this

src/workflows/chapters.ts
cjpillsbury: Will figure out a robust solution for this as a followup.

src/workflows/embeddings.ts
cjpillsbury: Ok looks like some of these were needed for vercel workflows, so we'll treat any addressing of these as a followup.

PR #60, opened 2025-12-11
fix: use workflow aws

cjpillsbury:

PR #61, opened 2025-12-11
fix: use workflow aws

dylanjha:
takes over #60 

- [x] ESM only build
- [x] load aws deps with dynamic imports because they are breaking if we don't do that


monsieurBoutte (approved):
LGTM

PR #66, opened 2025-12-12
fix: only support ambient env variables for credentials and related

cjpillsbury:


docs/WORKFLOWS.md
```diff
@@ -255,18 +255,8 @@ S3_ACCESS_KEY_ID=your-access-key
 S3_SECRET_ACCESS_KEY=your-secret-key
 ```
 
-Or pass directly:
+> **⚠️ Important:** For security reasons of possible use cases, we do not allow passing any credentials or other secure values via workflow options, including `S3_ACCESS_KEY_ID` or `S3_SECRET_ACCESS_KEY`
 
-```typescript
```
cjpillsbury, line 260: not sure how we want to phrase this one... I omitted for now, but happy to re-add with better framing/phrasing

docs/PRIMITIVES.md
```diff
@@ -504,15 +498,12 @@ export async function analyzeSentiment(
...
-const signingContext = {
-  signingKey: process.env.MUX_SIGNING_KEY,
-  privateKey: process.env.MUX_PRIVATE_KEY
-};
-
-// Primitives automatically sign URLs when context provided
-const storyboardUrl = await getStoryboardUrl(playbackId, 640, signingContext);
-const thumbnails = await getThumbnailUrls(playbackId, duration, { signingContext });
-const transcript = await fetchTranscriptForAsset(asset, playbackId, { signingContext });
+const shouldSign = !!(process.env.MUX_SIGNING_KEY && process.env.MUX_PRIVATE_KEY);
```
monsieurBoutte: ❓ is it possible to have these tokens and yet some assets don't require signed urls?

docs/WORKFLOWS.md
```diff
@@ -255,18 +255,8 @@ S3_ACCESS_KEY_ID=your-access-key
 S3_SECRET_ACCESS_KEY=your-secret-key
 ```
 
-Or pass directly:
+> **⚠️ Important:** For security reasons of possible use cases, we do not allow passing any credentials or other secure values via workflow options, including `S3_ACCESS_KEY_ID` or `S3_SECRET_ACCESS_KEY`
```
monsieurBoutte: 👍 good call out here

docs/PRIMITIVES.md
dylanjha: That's a very good catch. In fact, I think it's fairly normal to have a mix of public / private playback IDs

cjpillsbury: Yessir. It's the same reason we "under the hood" already had the concept of a `policy` that we checked in various places.

cjpillsbury: We treat them as equivalent in our workflows though.

docs/WORKFLOWS.md
dylanjha:
```suggestion
> **⚠️ Important:** Currently, the only way to set secrets is as ENV vars. The reason for this is to avoid leaking secrets into Workflow DevKit observability tooling. `S3_ACCESS_KEY_ID` and `S3_SECRET_ACCESS_KEY` are considered secrets. The other values for S3: `S3_ENDPOINT`, `S3_REGION` and `S3_BUCKET` are not considered secrets. They can be passed in explicitly at runtime.
```

cjpillsbury: @dylanjha I like this. Mind (code)word-smithing some version of the `typescript` codeblock demonstrating this?

docs/PRIMITIVES.md
monsieurBoutte: but i think the docs as you have it here might be confusing to the user right? Or it might imply the wrong message to them in that all assets should be signed if those keys are present

dylanjha: Oh, this is just a docs diff -- I missed the context, I thought this was in our SDK code, making an incorrect determination if we should sign a playback ID.

I see what the code sample is doing. I think maybe the code sample could be more clear `shouldSign: true`? Or maybe it is and I just got tripped up

docs/WORKFLOWS.md
dylanjha: `s3Endpoint`, `s3Region` and `s3Bucket` can be over-ridden

```ts
const result = await translateCaptions(assetId, "en", "es", {
  provider: "anthropic",
  s3Endpoint: "https://your-endpoint.com",
  s3Region: "auto",
  s3Bucket: "your-bucket",
})
```

docs/PRIMITIVES.md
cjpillsbury: nah I'll do that and add a code comment mentioning that the expectation is you have to have env vars defined

dylanjha (approved):

docs/PRIMITIVES.md
cjpillsbury: Updated!

PR #75, opened 2025-12-16
chore: And try to fix eval hosting once again.

philcluff:


monsieurBoutte:
taking an alternative path to publishing eval reports

PR #76, opened 2025-12-16
chore: update git workflow for publishing evals

monsieurBoutte:
# Overview

Fixes the Evalite UI hosted on GitHub Pages so that clicking on individual evals no longer shows "404: Eval Not Found".

# What was changed

### `.github/workflows/publish-evals-ui.yml`
- Added `--basePath=/ai/` to the export command so assets resolve correctly at `muxinc.github.io/ai/`
- Added a step to copy `index.html` to `404.html` for SPA client-side routing support

### `.gitignore`
- Added `.cursor/*` to ignore Cursor IDE files

# Suggested review order

1. `.github/workflows/publish-evals-ui.yml` — the main fix
2. `.gitignore` — unrelated housekeeping

# Root cause

GitHub Pages has two issues with Vite-based SPAs:
1. **Base path**: Assets must be served from `/repo-name/` not `/`
2. **SPA routing**: Direct navigation to `/suite/Burned-in%20Captions` returns 404 because that file doesn't exist — the `404.html` redirect lets the client-side router handle it

### DIFF
--- a/.github/workflows/publish-evals-ui.yml
+++ b/.github/workflows/publish-evals-ui.yml
@@ -49,7 +49,7 @@ jobs:
           S3_SECRET_ACCESS_KEY: ${{ secrets.S3_SECRET_ACCESS_KEY }}
         run: |
           npx evalite run
-          npx evalite export --output=./ui-output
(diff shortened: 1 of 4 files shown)


monsieurBoutte:
taking an alternative path to publishing eval reports

PR #83, opened 2025-12-19
chore: publish eval results

monsieurBoutte:
# Overview

Migrates evalite results publishing from GitHub Pages to an API-based ingest system. Instead of exporting a static UI to Pages, evals now run and POST raw JSON results to a configured endpoint (used by `evaluating-mux-ai`).

# What was changed

| File | Change |
|------|--------|
| `scripts/export-evalite-results.ts` | New script to run evalite and output `evalite-results.json` |
| `scripts/post-evalite-results.ts` | New script to POST results to API with repo/sha/ref metadata |
| `.github/workflows/publish-evalite-results.yml` | Replaced GitHub Pages deploy with API POST workflow |
| `src/env.ts` | Added `EVALITE_RESULTS_ENDPOINT` and `EVALITE_INGEST_SECRET` env vars |
| `package.json` | Added `evalite:post-results:dev` and `evalite:post-results:production` scripts |
| `docs/EVALS.md` | Updated CI/CD docs to reflect new flow |
| `.env.example` | Added new env var examples |
| `.gitignore` | Ignore `evalite-results.json` |

# Suggested review order

1. `scripts/post-evalite-results.ts` — Core logic: builds envelope with git metadata, POSTs to endpoint
2. `scripts/export-evalite-results.ts` — Simple runner that outputs JSON
3. `.github/workflows/publish-evalite-results.yml` — CI integration
4. `src/env.ts` — Env schema additions
5. `docs/EVALS.md` — Documentation updates

# Key details

- **Auth**: Uses `x-evalite-secret` header with `EVALITE_INGEST_SECRET`
- **Metadata**: Payload includes `repo`, `sha`, `ref`, `githubRunId`, `githubRunAttempt`
- **CLI flags**: `--dry-run` (-d) prints payload without posting; `--keep-file` (-k) skips deleting the JSON
- **Cleanup**: Results file deleted after successful POST (unless `--keep-file`)

### DIFF
--- a/.github/workflows/publish-evalite-results.yml
+++ b/.github/workflows/publish-evalite-results.yml
@@ -1,38 +1,28 @@
-name: Publish Evals UI
+name: Publish Evalite Results
 
 on:
   push:
     branches: [main]
   workflow_dispatch: # Allows manual triggering from any branch
 
-permissions:
-  contents: read
-  pages: write
-  id-token: write
-
-concurrency:
-  group: pages
-  cancel-in-progress: true
-
 jobs:
   build-and-deploy:
     runs-on: ubuntu-latest
-    environment:
-      name: github-pages
-      url: ${{ steps.deployment.outputs.page_url }}
     steps:
       - uses: actions/checkout@v4
 
       - uses: actions/setup-node@v4
         with:
-          node-version: "22"
+          node-version-file: .nvmrc
+          cache: npm
 
       - run: npm install
 
-      - name: Run Evalite and Export UI
+      - name: Run Evalite and Export Results
         env:
           MUX_TOKEN_ID: ${{ secrets.MUX_TOKEN_ID }}
           MUX_TOKEN_SECRET: ${{ secrets.MUX_TOKEN_SECRET }}
+          EVALITE_INGEST_SECRET: ${{ secrets.EVALITE_INGEST_SECRET }}
           MUX_SIGNING_KEY: ${{ secrets.MUX_SIGNING_KEY }}
           MUX_PRIVATE_KEY: ${{ secrets.MUX_PRIVATE_KEY }}
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
@@ -45,15 +35,10 @@ jobs:
           S3_BUCKET: ${{ secrets.S3_BUCKET }}
           S3_ACCESS_KEY_ID: ${{ secrets.S3_ACCESS_KEY_ID }}
           S3_SECRET_ACCESS_KEY: ${{ secrets.S3_SECRET_ACCESS_KEY }}
+          EVALITE_RESULTS_ENDPOINT: ${{ secrets.EVALITE_RESULTS_ENDPOINT }}
         run: |
-          npx evalite run
-          npx evalite export --output=./ui-output
+          npm run evalite:post-results:production
 
-      - name: Upload static UI for Pages
-        uses: actions/upload-pages-artifact@v3
-        with:
-          path: ui-output
-
-      - name: Deploy to GitHub Pages
-        id: deployment
-        uses: actions/deploy-pages@v4
+      - name: Cleanup Evalite Results File
+        if: always()
+        run: rm -f evalite-results.json
--- a/docs/EVALS.md
+++ b/docs/EVALS.md
@@ -67,17 +67,22 @@ npx evalite serve summarization.eval.ts
 
 ### CI/CD
 
-Evals run automatically on pushes to `main` (or via manual workflow dispatch). The CI job executes the evals, exports the json output, and posts the raw results to the Evalite API used by `evaluating-mux-ai`:
+Evals run automatically on pushes to `main` (or via manual workflow dispatch). The CI job executes the evals, exports the JSON output, and posts the raw results to the Evalite API used by `evaluating-mux-ai`.
+
+**For local development/testing:**
 
 ```bash
-# Run evals and write evalite-results.json
-npx tsx scripts/export-evalite-results.ts
+# Run evals and export results as a dry run (inspect without publishing)
+npm run evalite:post-results:dev
+```
 
-# Export static UI output (optional)
-npx evalite export --output ./ui-output
+**For production (internal maintainers only):**
 
-# Post results to the API (deletes evalite-results.json on success)
-npx tsx scripts/post-evalite-results.ts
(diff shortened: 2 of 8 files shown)


docs/EVALS.md
```diff
@@ -67,17 +67,20 @@ npx evalite serve summarization.eval.ts
 
 ### CI/CD
 
-Evals run automatically on every push to `main` and on pull requests:
+Evals run automatically on pushes to `main` (or via manual workflow dispatch). The CI job executes the evals, exports the json output, and posts the raw results to the Evalite API used by `evaluating-mux-ai`:
```
cjpillsbury: question(blocking): Do manual/local workflow dispatches also update the prod api? If so, let's make sure we don't noise up our "official" evals scores when doing testing/dev iteration cycles.

docs/EVALS.md
```diff
@@ -67,17 +67,20 @@ npx evalite serve summarization.eval.ts
...
 
 ```bash
-# Run evals once and exit (CI mode)
-npx evalite run
+# Run evals and write evalite-results.json
+npx tsx scripts/export-evalite-results.ts
```
cjpillsbury: nitpick(blocking): Unless I'm misunderstanding, these should refer to your newly added package.json scripts, no?

docs/EVALS.md
monsieurBoutte: good question!

only if they run the production script. Otherwise, the dev script is just for inspecting the results of a dry run

```
"evalite:post-results:dev": "npx tsx scripts/export-evalite-results.ts && npx tsx scripts/post-evalite-results.ts -d -k",
"evalite:post-results:production": "npx tsx scripts/export-evalite-results.ts && npx tsx scripts/post-evalite-results.ts",
```

monsieurBoutte: just running the regular evals and/or eval scripts locally won't affect any published evals

.github/workflows/publish-evalite-results.yml
```diff
@@ -45,15 +35,10 @@ jobs:
...
-        with:
-          path: ui-output
+          npm run evalite:post-results:production
 
-      - name: Deploy to GitHub Pages
-        id: deployment
-        uses: actions/deploy-pages@v4
+      - name: Cleanup Evalite Results File
+        if: always()
+        run: rm -f evalite-results.json
```
cjpillsbury, line 44: question(non-blocking): Isn't this redundant, since `post-evalite-results` will `unlink()` the file by default?

docs/EVALS.md
monsieurBoutte: yup, good catch

scripts/post-evalite-results.ts
```diff
@@ -0,0 +1,675 @@
...
+    generatedAt: string;
+    model?: {
+      provider?: string;
+      modelId?: string;
+    };
+    workflows: WorkflowInsightPayload[];
+  };
+}
+
+function execGit(command: string) {
```
cjpillsbury: nitpick(blocking): This just happens to be used for executing `git` commands but is really just a generic wrapper. Consider a more generic name, e.g. `execTextIOCmd` (I dunno. Names Are Hard™️)

.github/workflows/publish-evalite-results.yml
monsieurBoutte: I believe it's a helpful fail safe incase the script fails before reaching the unlink

scripts/post-evalite-results.ts
```diff
@@ -0,0 +1,675 @@
...
+  }
+}
+
+interface Options {
+  dryRun: boolean;
+  keepFile: boolean;
+}
+
+const WORKFLOW_MATCHERS = [
+  { key: "burned_in_captions", regex: /burned[- ]in[- ]captions/i },
```
cjpillsbury: thought(non-blocking): Figuring out how to parameterize these instead of hard-coding them would be good for maintainability. Def can be a future concern/consideration, though.

scripts/post-evalite-results.ts
```diff
@@ -0,0 +1,675 @@
...
+    }
+  }
+  return null;
+}
+
+function coerceNumber(value: unknown): number | undefined {
+  return typeof value === "number" && Number.isFinite(value) ? value : undefined;
+}
+
+function computeWorkflowStats(suite: EvaliteSuite, workflowKey: string): WorkflowInsightStats {
```
cjpillsbury, line 219: thought(non-blocking): Yeah there's a lot of coupling here re: code assumptions. Would be good to figure out how to abstract this stuff. Ideally the script for posting would be a lot simpler, which is what I assumed it would be before starting this review.

scripts/post-evalite-results.ts
```diff
@@ -0,0 +1,675 @@
...
+      undefined,
+    recommendations:
+      recommendations.quality || recommendations.latency || recommendations.expense ?
+        recommendations :
+        undefined,
+    notes: notes.length > 0 ? notes : undefined,
+  };
+}
+
+async function generateWorkflowInsights(suites: EvaliteSuite[]) {
```
cjpillsbury: issue(blocking): Since this is a public repo, this needs to be optional in its current implementation, since we cannot assume all users will have accounts/creds for Open AI.

cjpillsbury (changes requested):
A few minor blocking callouts and some bigger picture food for thought as followups, but overall 👌

scripts/post-evalite-results.ts
monsieurBoutte: Sure i can make that update. For what it's worth, this whole script is optional in the sense that it's only really useful for the internal maintainers of the project. This script only runs for us when main gets updated

monsieurBoutte: on second thought, this step is not optional as it helps with suggested recommendations based on the results of the evals. But what i can do is help support the other two providers (google or anthropic) as alternatives.

But again this is still largely an internal maintainer script. It can be adopted for someone whose forked the project by inspecting the output of a dry run but I don't see OSS contributors actually using this

cjpillsbury: cool worth a followup convo, but that works for now.

cjpillsbury (approved):
🚢 🇮🇹

scripts/post-evalite-results.ts
monsieurBoutte: good call out. Just added support for this

scripts/post-evalite-results.ts
monsieurBoutte: yeah fair point. Certainly open to simplifying this

Issue #51, opened 2025-12-08
Make it easier to run integration tests locally


dylanjha:
Right now running integration tests requires MUX_TOKEN_ID and MUX_TOKEN_SECRET for a specific Mux env (that owns the hard-coded asset IDs).

Maybe those hard-coded asset IDs should be moved to ENV vars -- or at least overridable with env vars, and then we can explain in the README if you want to run integration tests then you need to set your own MUX creds & asset IDs that exist in that env?

thisisayushverma:
hey i would like to pick this issues..

Faizan711:
Hi there!

Is there anyone still working on this?
if not, then i would like to pick it and up and get it done

monsieurBoutte:
Just pushed up a PR to address this. Thanks for the feedback and interest 👍

PR #86, opened 2025-12-29
chore: move test ids into env vars

monsieurBoutte:
# Overview

This branch makes the integration test suite **configurable for external contributors** by centralizing Mux test asset IDs behind env vars (with sane defaults) and documenting the expected `.env.test` setup.

addresses #51 

# What was changed

- **Integration tests now use a shared test-asset helper** instead of hard-coded asset IDs:
  - Added `tests/helpers/mux-test-assets.ts` exporting `muxTestAssets` (`assetId`, `violentAssetId`, `burnedInCaptionsAssetId`).
  - Updated integration tests to import `muxTestAssets` and reference its IDs.
  - Removed per-test `import "../../src/env";` side-effect imports in favor of importing the helper (which loads env via `src/env.ts`).

- **Env schema extended with test-only variables** (optional):
  - `src/env.ts` adds `MUX_TEST_ASSET_ID`, `MUX_TEST_ASSET_ID_VIOLENT`, `MUX_TEST_ASSET_ID_BURNED_IN_CAPTIONS`.

- **Docs + developer experience updates**:
  - Added `env.test.example` as the canonical template for `.env.test` (gitignored).
  - `CONTRIBUTING.md` updated to recommend `cp env.test.example .env.test` and explains required integration-test assets and env vars.
  - `README.md` adds a short note pointing to the above docs for integration tests.

# Suggested review order

1. `tests/helpers/mux-test-assets.ts` (new) — single source of truth for integration-test asset IDs + env var overrides.
2. `src/env.ts` — new optional schema entries used by the helper.
3. `tests/integration/*.test.ts` and `tests/integration/*.test.workflowdevkit.ts` — mechanical swaps from hard-coded IDs to `muxTestAssets.*`.
4. `env.test.example` — new example env file aligned with test scripts using `DOTENV_CONFIG_PATH=.env.test`.
5. `CONTRIBUTING.md` + `README.md` — documentation updates guiding contributors through the new flow.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -42,14 +42,15 @@ npm install
 ### Environment Configuration
 
 ```bash
-# Copy the example environment file
-cp .env.example .env
+# Copy the example test environment file
+cp env.test.example .env.test
 
-# Edit .env with your credentials
+# Edit .env.test with your credentials
 # You'll need:
 # - Mux credentials (MUX_TOKEN_ID, MUX_TOKEN_SECRET)
 # - At least one AI provider API key (OPENAI_API_KEY, ANTHROPIC_API_KEY, etc.)
 # - S3 credentials if testing translation workflows
+# - Mux test asset IDs (MUX_TEST_ASSET_ID, etc.) if running integration tests with your own Mux env
 ```
 
 > **💡 Tip:** You don't need all credentials—only those for the workflows you're testing.
@@ -131,6 +132,21 @@ npm run test:integration
 # Run LLM evaluation tests
 npm run test:eval
 ```
+
+### Integration test assets (Mux)
+
+Integration tests call real Mux + AI provider APIs, so they require:
+
+- **Mux credentials**: `MUX_TOKEN_ID`, `MUX_TOKEN_SECRET`
+- **Assets that exist in the same Mux environment as those credentials**
+
+By default, the repo’s integration tests use Mux-owned demo assets (hard-coded defaults). If you’re using your own Mux credentials, override the asset IDs via env vars in `.env.test`:
+
+- **`MUX_TEST_ASSET_ID`**: General-purpose asset used by most workflow integration tests (recommended: has an English transcript/captions track).
+- **`MUX_TEST_ASSET_ID_VIOLENT`**: Asset likely to score above the violence threshold in moderation tests.
+- **`MUX_TEST_ASSET_ID_BURNED_IN_CAPTIONS`**: Asset with clearly visible burned-in (hardcoded) captions.
+
+For the canonical list of env vars + what each test asset is expected to represent, see `tests/helpers/mux-test-assets.ts`.
 
 ### Writing Tests
 
--- a/src/env.ts
+++ b/src/env.ts
@@ -35,6 +35,7 @@ const EnvSchema = z.object({
     "Mux burned-in captions test asset id",
   ),
   MUX_TEST_ASSET_ID_AUDIO_ONLY: optionalString("Mux test asset ID for audio-only assets.", "Mux test asset id for audio-only assets for testing"),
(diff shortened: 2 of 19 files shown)


src/env.ts
```diff
@@ -27,13 +27,25 @@ const EnvSchema = z.object({
   MUX_SIGNING_KEY: optionalString("Mux signing key ID for signed playback URLs.", "Used to sign playback URLs"),
   MUX_PRIVATE_KEY: optionalString("Mux signing private key for signed playback URLs.", "Used to sign playback URLs"),
 
+  // Test-only helpers (used by this repo's integration tests)
+  MUX_TEST_ASSET_ID: optionalString("Mux asset ID used by integration tests.", "Mux test asset id"),
```
philcluff, line 31: Are there any restrictions on what this needs to contain? Content duration? Presumedly it needs to not be violent for tests to pass correctly too?

src/env.ts
```diff
@@ -27,13 +27,25 @@ const EnvSchema = z.object({
...
+  MUX_TEST_ASSET_ID_VIOLENT: optionalString("Mux violent asset ID used by integration tests.", "Mux violent test asset id"),
+  MUX_TEST_ASSET_ID_BURNED_IN_CAPTIONS: optionalString(
+    "Mux burned-in captions asset ID used by integration tests.",
```
philcluff, line 34: Debating if we should be more explicit here, or more pushy link to the readme with more details on what these need to contain. What do you think?

CONTRIBUTING.md
```diff
@@ -132,6 +133,21 @@ npm run test:integration
...
+
+Integration tests call real Mux + AI provider APIs, so they require:
+
+- **Mux credentials**: `MUX_TOKEN_ID`, `MUX_TOKEN_SECRET`
+- **Assets that exist in the same Mux environment as those credentials**
+
+By default, the repo’s integration tests use Mux-owned demo assets (hard-coded defaults). If you’re using your own Mux credentials, override the asset IDs via env vars in `.env.test`:
+
+- **`MUX_TEST_ASSET_ID`**: General-purpose asset used by most workflow integration tests (recommended: has an English transcript/captions track).
+- **`MUX_TEST_ASSET_ID_VIOLENT`**: Asset likely to score above the violence threshold in moderation tests.
```
philcluff, line 146: Maybe add the default threshold here?

CONTRIBUTING.md
```diff
@@ -132,6 +133,21 @@ npm run test:integration
...
+### Integration test assets (Mux)
+
+Integration tests call real Mux + AI provider APIs, so they require:
+
+- **Mux credentials**: `MUX_TOKEN_ID`, `MUX_TOKEN_SECRET`
+- **Assets that exist in the same Mux environment as those credentials**
+
+By default, the repo’s integration tests use Mux-owned demo assets (hard-coded defaults). If you’re using your own Mux credentials, override the asset IDs via env vars in `.env.test`:
+
+- **`MUX_TEST_ASSET_ID`**: General-purpose asset used by most workflow integration tests (recommended: has an English transcript/captions track).
```
philcluff, line 145: How general purpose do we think this is?

philcluff (approved):
Approved, feel free to merge, but some pondering about in the comments about how we can best communicate the requirements for each type.

src/env.ts
monsieurBoutte: we can be more explicit but i figured the `MUX_TEST_ASSET_ID_VIOLENT` would help hint that this shouldn't be.

src/env.ts
monsieurBoutte: nah, I think running the test will help communicate more than overly verbose variable names and comments might. I could be wrong but it seems okay given the test coverage / descriptions of the test themselves help communicate what these assets are being used for, testing wise

CONTRIBUTING.md
monsieurBoutte: relatively general considering todays usage but that may change at some point

CONTRIBUTING.md
monsieurBoutte: I'm going to skip adding this for now since it isn't really intended on being a configurable env var.

PR #87, opened 2026-01-06
fix: Fix moderation implementation to work with Vercel Workflow

philcluff:
The moderation workflow was failing in Vercel's workflow system due to how the moderation functions were structured. Both requestOpenAIModeration and requestHiveModeration defined their async moderation functions inline as closures marked with "use step". When these inline functions were passed to processConcurrently(), Vercel's workflow engine couldn't properly serialize them for execution as separate workflow steps, causing the moderation functionality to break.

### DIFF
--- a/src/workflows/moderation.ts
+++ b/src/workflows/moderation.ts
@@ -139,6 +139,55 @@ async function processConcurrently<T>(
   return results;
 }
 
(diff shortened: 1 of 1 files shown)


cjpillsbury (approved):
LGTM! 🚢 🇮🇹

PR #88, opened 2026-01-07
chore: evals should use env vars

monsieurBoutte:
# Overview

This just ensures our eval coverage also makes use of environment variables, contributors can override, as opposed to always eagerly using hardcoded values


philcluff (approved):
LGTM±

PR #91, opened 2026-01-08
feat: multi-tenant support

monsieurBoutte:
# Overview

Adds multi-tenant credential support for Workflow Dev Kit. Credentials can now be encrypted client-side and passed safely through workflow serialization boundaries, then decrypted inside steps on the execution host.

  **Problem:** Workflow Dev Kit serializes workflow inputs/outputs for observability. Passing plaintext API keys through `start()` would expose secrets in logs/traces.

  **Solution:** AES-256-GCM encryption wrapper that encrypts credentials before workflow invocation and decrypts them inside steps using a shared secret key (`MUX_AI_WORKFLOW_SECRET_KEY`). 

Implementation also uses Web Crypto APIs to avoid Node.js modules in workflow functions.

---

# How this works on Vercel (serverless-only setup)

There are two phases, both running in Vercel serverless functions:

  1. **Trigger phase** — your API route calls `start()` and passes **encrypted** credentials.
  2. **Execution phase** — the workflow steps run in a separate serverless invocation; this is the **execution host** where decryption happens.

Set `MUX_AI_WORKFLOW_SECRET_KEY` in your Vercel project so every serverless function can decrypt during step execution.

  ```ts
  import { start } from "workflow/api";
  import { encryptForWorkflow, getSummaryAndTags } from "@mux/ai/workflows";

// Trigger phase (API route): encrypt before calling start()
  const workflowKey = process.env.MUX_AI_WORKFLOW_SECRET_KEY!;
  const encryptedCredentials = await encryptForWorkflow(
    {
      muxTokenId: "mux-token-id",
      muxTokenSecret: "mux-token-secret",
      openaiApiKey: "openai-api-key",
    },
    workflowKey,
  );

  // Execution phase happens later in a separate invocation.
  // The workflow runtime decrypts inside steps using MUX_AI_WORKFLOW_SECRET_KEY.
  const run = await start(getSummaryAndTags, [
    "your-asset-id",
    { provider: "openai", credentials: encryptedCredentials },
  ]);
```
  Key point: ciphertext crosses workflow boundaries; plaintext only exists inside the step runtime.

---

  # What was changed

  ### New files
  | File | Purpose |
  |------|---------|
  | `src/lib/workflow-crypto.ts` | `encryptForWorkflow` / `decryptFromWorkflow` helpers (AES-256-GCM) with optional key ID (`kid`) for rotation |
  | `src/lib/workflow-credentials.ts` | Resolution chain: encrypted creds → provider hook → env fallback |
  | `tests/unit/workflow-crypto.test.ts` | Round-trip, wrong-key, malformed payload, and key ID tests |

  ### Modified files

  **Core:**
  - `src/index.ts` — Exports crypto helpers, `setWorkflowCredentialsProvider`, and version from `package.json`
  - `src/types.ts` — `WorkflowCredentials`, `WorkflowCredentialsInput`, `Encrypted<T>` types
  - `src/env.ts` — Mux creds now optional; added `MUX_AI_WORKFLOW_SECRET_KEY`; requires either Mux creds OR workflow key
  - `src/lib/client-factory.ts` — `getMuxCredentialsFromEnv` / `getApiKeyFromEnv` now async with credentials param
  - `src/lib/providers.ts` — Added `resolveLanguageModelConfig` / `resolveEmbeddingModelConfig`; model factories now async
  - `src/lib/url-signing.ts` — `signUrl` now resolves signing context from credentials
  - `src/lib/workflow-crypto.ts` — Uses Web Crypto + `TextEncoder/TextDecoder`; crypto helpers are async
  - `src/lib/image-download.ts` — Uses `Uint8Array` and Web base64 helpers (no `node:buffer`)
  - `src/workflows/moderation.ts` — Hive file source now uses `Uint8Array` buffers
  - `src/lib/workflow-credentials.ts` — Awaits async decrypt

  **Primitives:**
  - `src/primitives/storyboards.ts`, `thumbnails.ts`, `transcripts.ts` — Accept `credentials` for signed URL generation

  **Workflows (all 7):**
  - `summarization.ts`, `chapters.ts`, `burned-in-captions.ts`, `moderation.ts`, `embeddings.ts`, `translate-captions.ts`, `translate-audio.ts`
  - Each threads `options.credentials` to Mux API calls, AI provider instantiation, and URL signing

  **Docs:**
  - `README.md` — Multi-tenant encryption examples + provider hook usage
  - `docs/API.md`, `docs/WORKFLOWS.md` — Updated credential guidance

  # Suggested review order

  1. **`src/lib/workflow-crypto.ts`** — AES-256-GCM implementation, IV/tag handling, `kid` support (WebCrypto)
  2. **`src/lib/workflow-credentials.ts`** — Resolution chain logic (decrypt → provider → env)
  3. **`tests/unit/workflow-crypto.test.ts`** — Crypto edge cases and key rotation tests
  4. **`src/types.ts`** — Type definitions that flow through entire codebase
  5. **`src/lib/client-factory.ts` + `src/lib/providers.ts`** — Async conversion pattern
  6. **One workflow (e.g., `summarization.ts`)** — Credential threading pattern
  7. **`README.md`** — User-facing docs for encryption + provider hook

  # Key design decisions

  | Decision | Rationale |
  |----------|-----------|
  | Encryption at rest, decryption in steps | Secrets never appear in workflow I/O serialization |
  | Web Crypto (no Node modules) | Workflow functions disallow Node.js core modules |
  | Async crypto helpers | SubtleCrypto APIs are async |
  | Fallback chain (encrypted → provider → env) | Flexible for multi-tenant or single-tenant deployments |
  | Versioned payload `{ v: 1, alg, kid?, iv, tag, ciphertext }` | Supports future algorithm upgrades and key rotation |
  | Optional `kid` field | Enables key rotation without breaking existing payloads |
  | Async credential resolution | Providers can fetch from external secrets managers |
  | Backwards compatible | Existing env-var-only usage continues unchanged |

# Testing notes

  ```bash
  # Run crypto unit tests
  npm test -- workflow-crypto

  # Run all unit tests
  npm test
```

Crypto tests cover: round-trip encryption, wrong key rejection, malformed payloads, key ID presence/absence, and backwards compatibility.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -67,7 +67,7 @@ Analyzes video thumbnails for inappropriate content using OpenAI's Moderation AP
   - `maxRetryDelay?: number` - Maximum delay between retries in milliseconds (default: 10000)
   - `exponentialBackoff?: boolean` - Whether to use exponential backoff (default: true)
 
-All credentials (`MUX_TOKEN_ID`, `MUX_TOKEN_SECRET`, `OPENAI_API_KEY`, `HIVE_API_KEY`) are automatically read from environment variables.
+Credentials (`MUX_TOKEN_ID`, `MUX_TOKEN_SECRET`, `OPENAI_API_KEY`, `HIVE_API_KEY`) are read from environment variables by default. For Workflow Dev Kit multi-tenant usage, pass encrypted `credentials` in options (see README).
 
 **Returns:**
 
--- a/docs/ENCRYPTED-SECRETS.md
+++ b/docs/ENCRYPTED-SECRETS.md
@@ -13,6 +13,124 @@ There are two phases, both running in Vercel serverless functions:
 When running inside a workflow runtime (or when `MUX_AI_WORKFLOW_SECRET_KEY` is set),
 passing plaintext `credentials` will throw an error to prevent accidental exposure.
 
+## Creating a Workflow Secret Key
+
+The `MUX_AI_WORKFLOW_SECRET_KEY` must be a 256-bit (32-byte) key encoded as
+base64. This key is used for AES-256-GCM encryption of your credentials.
+
+### Using OpenSSL
+
+Generate a cryptographically secure key with OpenSSL:
+
+```bash
+openssl rand -base64 32
+```
+
+This outputs a 44-character base64 string (32 bytes + padding) you can use
+directly as your secret key.
+
+### Using Node.js
+
+Generate a key programmatically with Node.js:
+
+```javascript
+const crypto = require("crypto");
+const key = crypto.randomBytes(32).toString("base64");
+console.log(key);
+```
+
+Or as a one-liner:
+
+```bash
+node -e "console.log(require('crypto').randomBytes(32).toString('base64'))"
+```
+
+### Storing the Key
+
+Store the generated key as an environment variable:
+
+```bash
+# .env or environment configuration
+MUX_AI_WORKFLOW_SECRET_KEY=your_base64_encoded_key_here
+```
+
+Keep this key secure and never commit it to version control.
+
+## Rotating the Workflow Secret Key
+
+Key rotation is the process of replacing an existing encryption key with a new
+one. You should rotate keys periodically or immediately if you suspect a key
+may have been compromised.
+
+### Key ID (`kid`) Support
+
+The `encryptForWorkflow` function accepts an optional third parameter—a key ID
+(`kid`)—that is stored in plaintext alongside the encrypted payload. This
+enables zero-downtime key rotation by allowing the execution host to identify
+which key was used for encryption.
+
+```typescript
+// Encrypt with a key ID
+const encrypted = await encryptForWorkflow(
+  credentials,
+  env.MUX_AI_WORKFLOW_SECRET_KEY!,
+  "key-2026-01", // key ID stored in payload.kid
+);
+```
+
+Security notes:
+- The `kid` is stored unencrypted—don't put sensitive data in it
+- If an attacker modifies `kid`, decryption simply fails—the wrong key can't
+  decrypt the payload
+
+### Rotation Strategy
+
+The built-in credential resolution always reads from `MUX_AI_WORKFLOW_SECRET_KEY`.
+For most use cases, a simple key swap works well:
+
+1. **Generate a new key** using the methods above.
+2. **Update `MUX_AI_WORKFLOW_SECRET_KEY`** on the execution host.
+3. **Deploy the change**.
+
+Any workflows triggered after the deployment will use the new key. Workflows
+already in progress will complete because credentials are decrypted at step
+execution time—if a step already started with the old key, it continues with
+those decrypted credentials.
+
+### Advanced: Zero-Downtime Rotation with Key IDs
+
+For zero-downtime rotation, you can use the `kid` field to tag encrypted payloads
+with a key identifier, then implement custom decryption logic that looks up the
+correct key. This requires bypassing the built-in credential resolution.
+
+```typescript
+import { decryptFromWorkflow, encryptForWorkflow } from "@mux/ai/workflows";
+
+// Trigger phase: tag payload with the key ID
+const encrypted = await encryptForWorkflow(
+  credentials,
+  process.env.MUX_AI_WORKFLOW_KEY_V2!,
+  "v2", // stored in payload.kid
+);
+
+// Execution phase: look up key by ID before decrypting
+function getKeyById(kid: string | undefined): string {
+  const keys: Record<string, string | undefined> = {
+    v1: process.env.MUX_AI_WORKFLOW_KEY_V1,
+    v2: process.env.MUX_AI_WORKFLOW_KEY_V2,
+  };
+  const key = keys[kid ?? "v1"];
+  if (!key) throw new Error(`Unknown key ID: ${kid}`);
+  return key;
+}
+
+const key = getKeyById(encrypted.kid);
+const decrypted = await decryptFromWorkflow(encrypted, key);
+```
+
+This approach lets you keep both keys active during the transition, then remove
+the old key once all in-flight workflows have completed.
+
 ## Example: Summarization (default providers only)
 
 `getSummaryAndTags` accepts the default language-model providers (`openai`,
--- a/docs/ENCRYPTED-SECRETS.md
+++ b/docs/ENCRYPTED-SECRETS.md
@@ -10,6 +10,9 @@ There are two phases, both running in Vercel serverless functions:
 - Trigger phase — your API route calls `start()` and passes encrypted credentials.
 - Execution phase — the workflow steps run in a separate serverless invocation; this is the execution host where decryption happens.
 
+When running inside a workflow runtime (or when `MUX_AI_WORKFLOW_SECRET_KEY` is set),
+passing plaintext `credentials` will throw an error to prevent accidental exposure.
+
 ## Example: Summarization (default providers only)
 
 `getSummaryAndTags` accepts the default language-model providers (`openai`,
--- a/docs/ENCRYPTED-SECRETS.md
+++ b/docs/ENCRYPTED-SECRETS.md
@@ -1,0 +1,90 @@
+# Encrypted Secrets
+
+This guide shows how encrypted credentials work with workflows and how provider
+type coverage narrows to each workflow's supported providers.
+
+## How this works on Vercel (serverless-only setup)
+
+There are two phases, both running in Vercel serverless functions:
+
+- Trigger phase — your API route calls `start()` and passes encrypted credentials.
+- Execution phase — the workflow steps run in a separate serverless invocation; this is the execution host where decryption happens.
+
+## Example: Summarization (default providers only)
+
+`getSummaryAndTags` accepts the default language-model providers (`openai`,
+`anthropic`, `google`). The type of `provider` is restricted to those values.
+
+```typescript
+import { start } from "workflow/api";
+import { encryptForWorkflow, getSummaryAndTags } from "@mux/ai/workflows";
+
+// Trigger phase (API route): encrypt before calling start()
+const workflowKey = process.env.MUX_AI_WORKFLOW_SECRET_KEY!;
+const encryptedCredentials = await encryptForWorkflow(
+  {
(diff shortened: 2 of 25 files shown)


src/lib/client-factory.ts
```diff
@@ -6,54 +6,30 @@ import type {
...
-      elevenlabs: "ELEVENLABS_API_KEY",
-    };
-    throw new Error(
-      `${provider} API key is required. Set ${envVarNames[provider]} environment variable.`,
-    );
-  }
-
-  return apiKey;
+export async function getApiKeyFromEnv(
+  provider: "openai" | "anthropic" | "google" | "hive" | "elevenlabs",
```
mmcc: Should this list get exported from types? Feels like it should also probably be used [here](https://github.com/muxinc/ai/blob/main/src/types.ts#L97), for example? 

Looking further down in the PR, does seem like there's [`SupportedProvider`](https://github.com/muxinc/ai/blob/b622d9f0302734ce95ab40e5825797b6b7a1875a/src/lib/providers.ts#L10)?

src/lib/workflow-credentials.ts
```diff
@@ -0,0 +1,206 @@
...
+    throw new Error(
+      "Mux credentials are required. Provide encrypted workflow credentials or set MUX_TOKEN_ID and MUX_TOKEN_SECRET environment variables.",
+    );
+  }
+
+  return { muxTokenId, muxTokenSecret };
+}
+
+/** Supported AI/ML provider identifiers for API key resolution */
+type ProviderApiKey = "openai" | "anthropic" | "google" | "hive" | "elevenlabs";
```
mmcc: See above comment on shared type for this sucker.

src/lib/workflow-crypto.ts
```diff
@@ -0,0 +1,270 @@
...
+}
+
+function base64ToBytes(value: string, label: string): Uint8Array {
+  if (!value) {
+    throw new Error(`Invalid encrypted payload: missing ${label}.`);
+  }
+  const normalized = value.length % 4 === 0 ? value : value + "=".repeat(4 - (value.length % 4));
+  if (!BASE64_ALPHABET_RE.test(normalized)) {
+    throw new Error(`Invalid encrypted payload: ${label} is not base64.`);
+  }
```
mmcc: Running a regex on what could be a pretty big string here feels a little overly defensive to me given that `atob` is already going to throw an exception if it's not valid base64 (line 80).

src/lib/workflow-crypto.ts
```diff
@@ -0,0 +1,270 @@
...
+  if (typeof globalThis.atob !== "function") {
+    throw new TypeError("Base64 decoder is not available in this environment.");
+  }
```
mmcc: What JS runtime could we possibly be running in that doesn't support a basic runtime feature like this?

src/lib/workflow-crypto.ts
```diff
@@ -0,0 +1,270 @@
...
+}
+
+/** Converts key to bytes and validates it's exactly 32 bytes (256 bits) */
+function normalizeKey(key: Uint8Array | string): Uint8Array {
+  let keyBytes: Uint8Array;
+  if (typeof key === "string") {
+    try {
+      keyBytes = base64ToBytes(key, "key");
+    } catch {
+      throw new Error("Invalid workflow secret key. Expected 32-byte base64 value.");
```
mmcc: We throw a bunch of descriptive errors above, then catch and eat them and just return a generic error. Explicitly adding the context of the secret key to the error is helpful, but might still be worth exposing what's thrown within `base64ToBytes`

src/workflows/translate-audio.ts
```diff
@@ -374,25 +393,19 @@ export async function translateAudio(
   const s3AccessKeyId = env.S3_ACCESS_KEY_ID;
   const s3SecretAccessKey = env.S3_SECRET_ACCESS_KEY;
 
-  if (!elevenLabsKey) {
-    throw new Error("ElevenLabs API key is required. Provide elevenLabsApiKey in options or set ELEVENLABS_API_KEY environment variable.");
+  if (!isEncryptedPayload(credentials)) {
+    const elevenLabsKey = credentials?.elevenLabsApiKey ?? env.ELEVENLABS_API_KEY;
```
mmcc: I'm too far down in this PR now to scroll back up, forgive my laziness here...but doesn't the credentials provider pass through credentials from the ENV too if they're there? Do we really need to check both everywhere as well?

mmcc (commented):
Looks really cool, and we should definitely make sure to get a second set of eyes on the stuff that's touching encryption/decryption from a security angle (not because I see something alarming, but because I don't feel confident in my ability to review that well).

src/lib/client-factory.ts
monsieurBoutte: yeah good catch

src/lib/workflow-crypto.ts
monsieurBoutte: if i recall correctly it was a wrinkle at vercels workflow runtime

src/env.ts
```diff
@@ -21,8 +22,12 @@ function requiredString(description: string, message?: string) {
 const EnvSchema = z.object({
   NODE_ENV: z.string().default("development").describe("Runtime environment."),
 
-  MUX_TOKEN_ID: requiredString("Mux access token ID.", "Required to access Mux APIs"),
-  MUX_TOKEN_SECRET: requiredString("Mux access token secret.", "Required to access Mux APIs"),
+  MUX_TOKEN_ID: optionalString("Mux access token ID.", "Required to access Mux APIs"),
+  MUX_TOKEN_SECRET: optionalString("Mux access token secret.", "Required to access Mux APIs"),
+  MUX_AI_WORKFLOW_SECRET_KEY: optionalString(
```
daniel-hayes, line 27: Should we include some docs that go through the process of creating a new key, and how to rotate the key? Using `openssl` or Node.js etc. 

Not necessary for this PR, but having a runbook on how we handle key rotation would be great too.

docs/API.md
```diff
@@ -67,7 +67,7 @@ Analyzes video thumbnails for inappropriate content using OpenAI's Moderation AP
   - `maxRetryDelay?: number` - Maximum delay between retries in milliseconds (default: 10000)
   - `exponentialBackoff?: boolean` - Whether to use exponential backoff (default: true)
 
-All credentials (`MUX_TOKEN_ID`, `MUX_TOKEN_SECRET`, `OPENAI_API_KEY`, `HIVE_API_KEY`) are automatically read from environment variables.
+Credentials (`MUX_TOKEN_ID`, `MUX_TOKEN_SECRET`, `OPENAI_API_KEY`, `HIVE_API_KEY`) are read from environment variables by default. For Workflow Dev Kit multi-tenant usage, pass encrypted `credentials` in options (see README).
```
daniel-hayes, line 70: Is there a way to detect if this code is being run in a workflow automatically? I'm wondering if we can enforce needing to pass encrypted credentials automatically based on the runtime _or_ if a  `MUX_AI_WORKFLOW_SECRET_KEY` var is set. I think documenting this is also fine for now, but if Vercel exposes workflow env variables by default, this could be a nice way to reduce possibility of accidental exposure.

And if we just check for `MUX_AI_WORKFLOW_SECRET_KEY`, we prevent needing to explicitly pass the secret into `encryptForWorkflow`. Open to discussing the tradeoffs though!

src/lib/workflow-credentials.ts
```diff
@@ -0,0 +1,206 @@
...
+ *
+ * Resolution order (later sources override earlier):
+ * 1. Credentials from the registered provider
+ * 2. Decrypted credentials (if input is an encrypted payload)
+ *    OR plain credentials object (if input is already decrypted)
+ *
+ * @param credentials - Optional credentials input (encrypted or plain object)
+ * @returns Merged credentials object
+ */
+export async function resolveWorkflowCredentials(
```
daniel-hayes, line 103: should we reject plaintext payloads if the MUX_AI_WORKFLOW_SECRET_KEY is an env variable is set?

src/lib/workflow-crypto.ts
daniel-hayes: good callout - would having a size/length limit suffice?

daniel-hayes (commented):
Looking really good!

src/env.ts
monsieurBoutte: yeah i think that's a good idea

docs/API.md
monsieurBoutte: that's a great callout. I know there is indeed a way of figuring out whether or not your in a workflow environment. I'll have to confirm with the Vercel team

src/lib/workflow-credentials.ts
monsieurBoutte: yes! Just added support for that

src/lib/workflow-crypto.ts
monsieurBoutte: opt'ed for removing the regex check here

src/lib/workflow-crypto.ts
monsieurBoutte: removed 👍

src/workflows/translate-audio.ts
monsieurBoutte: yup, good spot

src/env.ts
monsieurBoutte: added 👍

docs/API.md
monsieurBoutte: added 👍

docs/ENCRYPTED-SECRETS.md
daniel-hayes, line 1: 🔥

src/lib/workflow-credentials.ts
```diff
@@ -0,0 +1,239 @@
...
+ */
+export function setWorkflowCredentialsProvider(provider?: WorkflowCredentialsProvider): void {
+  workflowCredentialsProvider = provider;
+}
+
+/**
+ * Detects whether code is running inside a Workflow Dev Kit runtime.
+ * getWorkflowMetadata throws when invoked outside a workflow.
+ */
+async function isWorkflowRuntime(): Promise<boolean> {
```
daniel-hayes, line 39: awesome

daniel-hayes (approved):
Looks good!

PR #95, opened 2026-01-23
chore: mark workflow external in tsup

monsieurBoutte:
# Overview

## Issue

there was an error spotted to the published @mux/ai bundle: where it’s currently bundling the workflow dev‑kit (pulled in via the dynamic `import("workflow")` check in `workflow-credentials.ts`). That brings in @workflow/core and devalue, and the workflow sandbox runs without a global require, so the bundled CJS require("async_hooks") explodes with “require is not defined” as this is intended on being an ESM only package.

## What changed

fixed the build by externalizing workflow, so it no longer gets bundled into @mux/ai and the output is pure ESM without require.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,13 +46,13 @@
       }
     },
     "node_modules/@ai-sdk/anthropic": {
-      "version": "3.0.16",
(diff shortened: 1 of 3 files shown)


monsieurBoutte:
closing in favor of #96

PR #96, opened 2026-01-23
chore: updating tsup

monsieurBoutte:
# Overview

This branch adds shared agent skills, and makes workflow encryption/base64 utilities work in more environments. It also updates encrypted secrets docs, build externals, and runtime dependencies to support the new crypto and schema tooling.

# What was changed

- Added skill definitions under `.agents/skills/` with symlinks in `.claude/`, `.codex/`, and `.github/`.
- Swapped workflow crypto to use `@noble/ciphers` AES-GCM and added base64 fallbacks when `atob`/`btoa` are unavailable.
- Improved error detail handling in workflow credential decryption.
- Updated encrypted secrets docs imports to use `@mux/ai` for encryption helpers.
- Added new externals (`@noble/ciphers`, `workflow`) in `tsup`.

# Suggested review order

1. `src/lib/workflow-crypto.ts` (crypto/base64 changes, noble AES-GCM)
2. `src/workflows/summarization.ts` (schema/output handling, prompt tweaks, usage guards)
3. `src/lib/workflow-credentials.ts` (error detail handling)
4. `package.json` (new dep for noble)
5. `docs/ENCRYPTED-SECRETS.md` (import adjustments)
6. `tsup.config.ts` (external list)
7. `.agents/skills/**` and symlinked skill folders (new documentation + symlinks)

### DIFF
--- a/src/workflows/summarization.ts
+++ b/src/workflows/summarization.ts
@@ -1,7 +1,6 @@
-import { valibotSchema } from "@ai-sdk/valibot";
 import { generateText, Output } from "ai";
 import dedent from "dedent";
-import * as v from "valibot";
+import { z } from "zod";
 
 import type { ImageDownloadOptions } from "@mux/ai/lib/image-download";
 import { downloadImageAsBase64 } from "@mux/ai/lib/image-download";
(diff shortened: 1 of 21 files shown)


joshmux (approved):

src/workflows/summarization.ts
```diff
@@ -396,8 +404,16 @@ async function analyzeStoryboard(
     ],
   });
 
+  console.warn("[summarization] response:", JSON.stringify(response, null, 2));
```
philcluff: I think this is some debugging left in that we should remove, or at least change the log verbosity of.

src/workflows/summarization.ts
```diff
@@ -433,8 +449,16 @@ async function analyzeAudioOnly(
...
```
philcluff: Same - should this be removed or level changed?

philcluff (approved):
Approved with two comments about logging / log levels. 

I don't like how much crypto has come in here to make serialisation work, buuuuut, assuming we can move to helpers from ai-sdk over time, this is fine.

src/workflows/summarization.ts
monsieurBoutte: oh yeah, i meant to remove those

monsieurBoutte:
> Approved with two comments about logging / log levels.
> 
> I don't like how much crypto has come in here to make serialisation work, buuuuut, assuming we can move to helpers from ai-sdk over time, this is fine.

agreed. And we should be able to do so as soon as we move over to the [native encryption they're adding for workflows](https://github.com/vercel/workflow/pull/762).

PR #99, opened 2026-02-03
chore: add asset durations for posted eval runs

monsieurBoutte:
# Overview

This branch expands evalite insights generation to include Ask Questions and adds asset duration metrics so we can track estimated cost per workflow we're shipping.

# What was changed

- Added Ask Questions to the workflow matcher so its eval results get summarized in insights.
- Added Mux asset duration lookups, caching, and new duration totals/missing counts in workflow stats.
- Threaded the duration cache through insight generation and surfaced new fields in the stats payload.
- Threaded asset duration and thumbnail count through usage stats returned for each workflow
- Documented how `usage.metadata` is populated by workflows and consumed in eval insights

# Notes on usage metadata

`usage.metadata` is populated by each workflow when it builds its response. Workflows compute
context like asset duration (and thumbnail count for thumbnail-based workflows), then attach
those values on the `usage` object returned by the workflow. The eval output captures that
workflow response, so `post-evalite-results.ts` can read `output.usage.metadata` to aggregate
per-case duration and thumbnail metrics without extra Mux API calls.

# Suggested review order

1. `scripts/post-evalite-results.ts`
2. `src/lib/mux-assets.ts`
3. `src/types.ts`
4. `src/workflows/ask-questions.ts`
5. `src/workflows/burned-in-captions.ts`
6. `src/workflows/chapters.ts`
7. `src/workflows/embeddings.ts`
8. `src/workflows/moderation.ts`
9. `src/workflows/summarization.ts`
10. `src/workflows/translate-audio.ts`
11. `src/workflows/translate-captions.ts`
12. `examples/ask-questions/basic-example.ts`
13. `examples/moderation/basic-example.ts`

### DIFF
--- a/src/lib/mux-assets.ts
+++ b/src/lib/mux-assets.ts
@@ -68,6 +68,10 @@ export async function getAssetDurationSeconds(
   });
 
   const asset = await mux.video.assets.retrieve(assetId);
+  return getAssetDurationSecondsFromAsset(asset);
+}
+
+export function getAssetDurationSecondsFromAsset(asset: MuxAsset): number | undefined {
   const duration = asset.duration;
   return typeof duration === "number" && Number.isFinite(duration) ? duration : undefined;
 }
--- a/src/lib/mux-assets.ts
+++ b/src/lib/mux-assets.ts
@@ -55,3 +55,19 @@ export async function getPlaybackIdForAsset(
 
   return { asset, playbackId, policy };
 }
+
+export async function getAssetDurationSeconds(
+  assetId: string,
+  credentials?: WorkflowCredentialsInput,
+): Promise<number | undefined> {
+  "use step";
+  const { muxTokenId, muxTokenSecret } = await getMuxCredentialsFromEnv(credentials);
+  const mux = new Mux({
+    tokenId: muxTokenId,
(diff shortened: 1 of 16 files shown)


src/lib/mux-assets.ts
```diff
@@ -55,3 +55,23 @@ export async function getPlaybackIdForAsset(
...
+  credentials?: WorkflowCredentialsInput,
+): Promise<number | undefined> {
+  "use step";
+  const { muxTokenId, muxTokenSecret } = await getMuxCredentialsFromEnv(credentials);
+  const mux = new Mux({
+    tokenId: muxTokenId,
+    tokenSecret: muxTokenSecret,
+  });
+
+  const asset = await mux.video.assets.retrieve(assetId);
```
philcluff: In what use cases do we need to get the duration without getting playback IDs?

Would it be better to have a single `getMuxAsset` call in the helpers, which makes the Mux Video API call, and then make sure that the playback ID and duration calls have to happen after that call?

I'm just worried that the current structure makes it a little too easy to double hit the video API when using the SDK...

philcluff (approved):
Looks great!

PR #101, opened 2026-02-06
chore: replace @aws-sdk dependencies

monsieurBoutte:
# Overview

Branch: `vb/replace-aws-sdks` (diffed against `main`).

This branch replaces SDK-specific workflow credential wrappers with serializable, plain-object credentials plus encryption support for Workflow DevKit boundaries. It also introduces a pluggable `storageAdapter` path while keeping an internal SigV4 default, and updates workflows/docs/tests accordingly.

# What was changed

- **Workflow credentials + encryption**
  - Added `encryptForWorkflow` / `decryptFromWorkflow` utilities and encrypted payload typing.
  - Refactored credential resolution to prioritize provider + encrypted input, and reject plaintext credentials in workflow runtimes.
  - Removed legacy workflow credential wrapper classes/files in favor of serializable credential objects.

- **Storage abstraction and default SigV4 path**
  - Added internal `s3-sigv4` implementation (PUT + presigned GET) with endpoint hardening (`https` enforcement and optional host allowlist).
  - Added `StorageAdapter` types and adapter helpers (`putObjectWithStorageAdapter`, `createPresignedGetUrlWithStorageAdapter`).
  - Added `WorkflowStorageClient` + workflow serialization registration for workflow-safe storage client transport.

- **Workflow integration updates**
  - Updated workflows (notably `translate-captions` and `translate-audio`) to use storage adapter plumbing and new credential resolution paths.
  - Updated provider/client helpers and related modules to consume new credential shapes.

- **Public API + env surface**
  - Exported new workflow crypto/storage utilities from `src/index.ts`.
  - Updated shared types (`WorkflowCredentials`, `WorkflowCredentialsInput`, `StorageAdapter`, etc.).
  - Extended env schema with workflow secret key and S3 endpoint allowlist support.

- **Docs/examples/tests**
  - Added `docs/WORKFLOW-ENCRYPTION.md` and `docs/STORAGE-ADAPTERS.md`.
  - Updated `README.md`, `docs/API.md`, and `docs/EXAMPLES.md`.
  - Added AWS SDK adapter example for caption translation.
  - Added unit coverage for SigV4 and workflow crypto; removed old provider-client wrapper tests.

- **Dependency/build churn**
  - `package.json` / lockfiles updated (including `@noble/ciphers`, workflow version bump, example deps).
  - Large lockfile diffs are mostly dependency graph noise; review functional code first.

# Suggested review order

1. `src/lib/workflow-credentials.ts`, `src/lib/workflow-crypto.ts`, `src/types.ts`
   Verify credential precedence, encryption enforcement, and type-level API changes.
2. `src/lib/s3-sigv4.ts`, `src/lib/storage-adapter.ts`, `src/lib/workflow-storage-client.ts`, `src/lib/workflow-serialization.ts`
   Validate storage behavior, signing correctness, and workflow-safe serialization.
3. Workflow call sites: `src/workflows/translate-captions.ts`, `src/workflows/translate-audio.ts` (+ other workflow touch points)
   Confirm adapter/credential integration and runtime behavior.
4. Surface exports/config: `src/index.ts`, `src/env.ts`, related helper updates (`src/lib/providers.ts`, `src/lib/client-factory.ts`)
   Check external API compatibility and env requirements.
5. Tests: `tests/unit/s3-sigv4.test.ts`, `tests/unit/workflow-crypto.test.ts`
   Ensure key paths are covered and assertions match intended behavior.
6. Docs/examples and finally lockfiles
   Validate user-facing guidance once implementation is approved.


monsieurBoutte:
~pausing this -- there might be a simpler path forward~

after further investigation, I'm convinced this is the most straightforward path. For bonus points though, I added optional storage adapters for downstream consumers to have some optionality at their disposal.

docs/STORAGE-ADAPTERS.md
```diff
@@ -0,0 +1,228 @@
+# Storage Adapters
+
+Use `storageAdapter` when you want custom object storage behavior (for example, runtime-specific SDK usage) instead of the built-in SigV4 uploader.
+
+## Why built-in `s3-sigv4` exists
```
philcluff, line 5: Love this doc, it really helps readers understand the intent well.

src/lib/s3-sigv4.ts
```diff
@@ -0,0 +1,296 @@
...
+
+  const response = await fetch(requestUrl, {
+    method: "PUT",
+    headers: {
+      "Authorization": authorization,
+      "x-amz-content-sha256": payloadHash,
+      "x-amz-date": amzDate,
+      ...(contentType ? { "content-type": contentType } : {}),
+    },
+    body: typeof body === "string" ? body : body,
```
philcluff: This ternary always returns `body` :)

src/lib/s3-sigv4.ts
```diff
@@ -0,0 +1,296 @@
...
+  const authorization = `${AWS4_ALGORITHM} Credential=${accessKeyId}/${credentialScope}, SignedHeaders=${signedHeaders}, Signature=${signature}`;
+  const requestUrl = `${resolvedEndpoint.origin}${canonicalUri}`;
+
+  const response = await fetch(requestUrl, {
+    method: "PUT",
+    headers: {
+      "Authorization": authorization,
+      "x-amz-content-sha256": payloadHash,
+      "x-amz-date": amzDate,
+      ...(contentType ? { "content-type": contentType } : {}),
```
philcluff: (Claude flagged this one, I would have missed it) 

Should `content-type` also be in `canonicalHeaders` or `signedHeaders` ?

A quick dig suggests that it might be optional for S3, but required on other providers?

src/lib/s3-sigv4.ts
philcluff, line 1: While I believe this works, I think we need some unit test coverage for this implementation, we could have subtle bugs in here that we could easily miss or regress without.

src/workflows/translate-audio.ts
```diff
@@ -263,59 +275,44 @@ async function uploadDubbedAudioToS3({
...
-      accessKeyId: s3AccessKeyId,
-      secretAccessKey: s3SecretAccessKey,
-    },
-    forcePathStyle: true,
-  });
+  const s3AccessKeyId = env.S3_ACCESS_KEY_ID;
+  const s3SecretAccessKey = env.S3_SECRET_ACCESS_KEY;
 
   // Create unique key for the audio file
   const audioKey = `audio-translations/${assetId}/auto-to-${toLanguageCode}-${Date.now()}.m4a`;
```
philcluff, line 293: I know you didn't write this (I did), but, we should probably throw a workflow ID (if we're running in workflow), or uuid in the path (as a fallback) rather than just relying on the date to uniquify this.

src/workflows/translate-captions.ts
```diff
@@ -145,54 +177,36 @@ async function uploadVttToS3({
...
-      accessKeyId: s3AccessKeyId,
-      secretAccessKey: s3SecretAccessKey,
-    },
-    forcePathStyle: true,
-  });
+  const s3AccessKeyId = env.S3_ACCESS_KEY_ID;
+  const s3SecretAccessKey = env.S3_SECRET_ACCESS_KEY;
 
   // Create unique key for the VTT file
   const vttKey = `translations/${assetId}/${fromLanguageCode}-to-${toLanguageCode}-${Date.now()}.vtt`;
```
philcluff, line 170: Same comment on paths.

philcluff (changes requested):
Looks great overall, a few minor comments inline.

The blocker here for me is the lack of test coverage on `s3-sigv4.ts` - let's at least get something minimal in place here before we land this one.

Also looks like the integration tests are timing out because the workflow queueing is all failing with:

```
    `  "message": "Package subpath './lib/s3-sigv4' is not defined by \\"exports\\" in /home/runner/work/ai/ai/package.json imported from /home/runner/work/ai/ai/src/lib/workflow-storage-client.ts",\n` +
```

src/workflows/translate-audio.ts
monsieurBoutte: discussed following up with 

```ts
import { monotonicFactory } from "ulid";

const createUlid = monotonicFactory();

const jobId = createUlid();
```

in another PR

monsieurBoutte: monotonic ULID gives IDs that are:

- Unique
- Time-ordered lexicographically (newer IDs sort after older ones)
- Still strictly increasing even within the same millisecond (monotonicFactory behavior)
- If jobs are listed/sorted by id, you naturally get creation order without a separate timestamp sort.
- It’s friendlier for logs/debugging because the ID encodes time.

src/lib/s3-sigv4.ts
monsieurBoutte: nice spot

monsieurBoutte: addressed in a follow up commit

src/lib/s3-sigv4.ts
```diff
@@ -0,0 +1,299 @@
...
+  shortDate: string,
+  region: string,
+): Promise<Uint8Array> {
+  const kDate = await hmacSha256Raw(textEncoder.encode(`AWS4${secretAccessKey}`), shortDate);
+  const kRegion = await hmacSha256Raw(kDate, region);
+  const kService = await hmacSha256Raw(kRegion, AWS4_SERVICE);
+  return hmacSha256Raw(kService, AWS4_REQUEST_TERMINATOR);
+}
+
+function formatAmzDate(date = new Date()): { amzDate: string; shortDate: string } {
```
walkerfrankenberg, line 84: nit: could be simpler to define a formatter and then just pass in the date:

```
const parts = new Intl.DateTimeFormat("en-US", {
    timeZone: "UTC",
    year: "numeric",
    month: "2-digit",
    day: "2-digit",
    hour: "2-digit",
    minute: "2-digit",
    second: "2-digit",
    hour12: false,
  })

function formatAmzDate(date = new Date()): { amzDate: string; shortDate: string } {
  parts.formatToParts(date);

  const map = Object.fromEntries(parts.map(p => [p.type, p.value]));

  const shortDate = `${map.year}${map.month}${map.day}`;
  const amzDate = `${shortDate}T${map.hour}${map.minute}${map.second}Z`;

  return { amzDate, shortDate };
}
```

or using ISO strings
```
function formatAmzDate(date = new Date()): { amzDate: string; shortDate: string } {
  const iso = date.toISOString(); // always UTC
  const shortDate = iso.slice(0, 10).replace(/-/g, "");
  const amzDate = iso.slice(0, 19).replace(/[-:]/g, "").replace(".000", "") + "Z";
  return { amzDate, shortDate };
}
```

src/lib/workflow-crypto.ts
```diff
@@ -0,0 +1,322 @@
...
+  const cleaned = value
+    .replace(/\s+/g, "")
+    .replace(/-/g, "+")
+    .replace(/_/g, "/");
+
+  if (!cleaned) {
+    return cleaned;
+  }
+
+  return cleaned.length % 4 === 0 ? cleaned : cleaned + "=".repeat(4 - (cleaned.length % 4));
```
walkerfrankenberg: nit: could we
```suggestion
  return cleaned?.length % 4 === 0 ? cleaned : cleaned + "=".repeat(4 - (cleaned.length % 4));
```

walkerfrankenberg (approved):
couple small things but overall looks good!

philcluff (approved):
Approved. 

Before merging, let's get a build of this branch into Stream.new, and prove it's working as expected.

PR #102, opened 2026-02-06
chore: swap out encryption strategy

monsieurBoutte:
# Overview

Replace the custom AES-256-GCM encryption layer (`workflow-crypto`) with Workflow DevKit's native class serialization system. Credentials are now transported across step boundaries using serializable client wrappers (`WorkflowMuxClient`, `WorkflowOpenAIClient`, etc.) instead of manually encrypting/decrypting a flat credentials object. This also removes the `@noble/ciphers` dependency and the `MUX_AI_WORKFLOW_SECRET_KEY` requirement.

# What was changed

### Encryption removed, native serialization added
- Deleted `src/lib/workflow-crypto.ts` (AES-256-GCM encrypt/decrypt) and its tests.
- Deleted `docs/ENCRYPTED-SECRETS.md` (no longer relevant).
- Removed `@noble/ciphers` from dependencies.
- Upgraded `workflow` to `^4.1.0-beta.53` (required for class-registry serialization support).

### New serializable credential wrappers
- **`src/lib/workflow-mux-client.ts`** — `WorkflowMuxClient` wraps Mux API credentials with `WORKFLOW_SERIALIZE` / `WORKFLOW_DESERIALIZE` hooks.
- **`src/lib/workflow-provider-clients.ts`** — `WorkflowOpenAIClient`, `WorkflowAnthropicClient`, `WorkflowGoogleClient`, `WorkflowHiveClient`, `WorkflowElevenLabsClient` — same pattern per provider.
- **`src/lib/workflow-native-credentials.ts`** — `WorkflowNativeCredentials` container + `serializeForWorkflow()` helper that bundles a Mux client and a provider client into one serializable unit.

### Credential resolution updated
- `src/lib/workflow-credentials.ts` — stripped out encrypt/decrypt paths; now resolves credentials from `WorkflowNativeCredentials`, the credentials-provider callback, or plain objects (auto-wrapped).
- `src/lib/client-factory.ts` — uses the new `resolveMuxClient()` / `resolveProviderApiKey()` helpers.

### Workflows updated
- `src/workflows/summarization.ts`, `moderation.ts`, `translate-audio.ts` — all updated to normalize credentials through `serializeForWorkflow()` and the new client wrappers.

### Exports & types
- `src/index.ts` — exports all new wrapper classes and `serializeForWorkflow`.
- `src/types.ts` — `WorkflowCredentials` / `WorkflowCredentialsInput` now reference the new client types.

### esbuild / Nitro bundling fixes
- Added `tsup.config.ts` with a post-build transform to name anonymous class expressions in dist output (esbuild strips class names; Nitro's bundler needs them for the class registry).
- Removed manual class registration that conflicted with Nitro's auto-generated `registerSerializationClass` calls.
- Replaced TypeScript parameter properties (`constructor(private readonly ...)`) with explicit field declarations in all workflow wrapper classes. Nitro's worker uses Node's strip-only TS mode, which cannot transform parameter properties.

### Docs
- `README.md` — rewrote the multi-tenant credentials section to show the new `serializeForWorkflow` + `setWorkflowCredentialsProvider` API; removed all encryption references.

# Suggested review order

1. **`src/lib/workflow-native-credentials.ts`** — small file, establishes the new serialization pattern.
2. **`src/lib/workflow-mux-client.ts`** — same pattern applied to the Mux client.
3. **`src/lib/workflow-provider-clients.ts`** — same pattern for every AI provider; review one class then skim the rest.
4. **`src/lib/workflow-credentials.ts`** — the core resolution logic that ties it all together.
5. **`src/lib/client-factory.ts`** — thin adapter, quick read.
6. **`src/workflows/summarization.ts`** → `moderation.ts` → `translate-audio.ts` — see how the new credential flow is consumed.
7. **`tsup.config.ts`** — post-build fix for class-name preservation.
8. **`src/index.ts`** / `src/types.ts` — public API surface changes.
9. **`README.md`** — verify the docs match the new API.


src/lib/workflow-credentials.ts
```diff
@@ -111,61 +90,122 @@ export async function resolveWorkflowCredentials(
...
-  const muxTokenId = resolved.muxTokenId ?? env.MUX_TOKEN_ID;
-  const muxTokenSecret = resolved.muxTokenSecret ?? env.MUX_TOKEN_SECRET;
+  // Prefer a pre-built muxClient from credentials
+  if (resolved.muxClient) {
+    return resolved.muxClient;
+  }
+
+  // Fall back to environment variables
+  const muxTokenId = env.MUX_TOKEN_ID;
+  const muxTokenSecret = env.MUX_TOKEN_SECRET;
```
daniel-hayes, line 125: Would we ever want to do this? Going back to the conversation we had earlier, perhaps it is better to only ever accept these values per request.

monsieurBoutte: I need you and @jsanford8 to catch me up on the recommended approach here. For context I was just matching the public docs on how our API handles Auth but now that there's an alternative auth approach, I'm sure there's room to simplify

daniel-hayes: Okay github being down spammed this message... I also wound up deleting this comment because I needed to think on this more. As a user, outside of workflows, using the SDK, it makes sense to set the token as an env variable.

daniel-hayes (approved):
Looks great - ignore my comment earlier, we can address any env variable discussions as a followup

src/lib/mux-assets.ts
```diff
@@ -55,6 +54,16 @@ export async function getPlaybackIdForAsset(
   return { asset, playbackId, policy };
 }
 
+export async function getPlaybackIdForAssetWithClient(
```
philcluff, line 60: Nit: There's a little re-use we could do here with getPlaybackIdForAsset, but I'm not sure it's worth it.

src/lib/workflow-credentials.ts
```diff
@@ -111,61 +90,122 @@ export async function resolveWorkflowCredentials(
...
+): Promise<WorkflowHiveClient | undefined>;
+export async function resolveProviderClient(
+  provider: "elevenlabs",
+  credentials?: WorkflowCredentialsInput,
+): Promise<WorkflowElevenLabsClient | undefined>;
+export async function resolveProviderClient<P extends ProviderClientProvider>(
+  provider: P,
+  credentials?: WorkflowCredentialsInput,
+): Promise<ProviderClientByProvider[P] | undefined> {
+  const resolved = await resolveWorkflowCredentials(credentials);
```
philcluff, line 257: Does this end up getting called twice for each instantiation? `resolveProviderClient` and `resolveProviderApiKey`? Does that matter?

philcluff (approved):
Approved with one nit and one question!

src/lib/workflow-credentials.ts
monsieurBoutte: great spot, yeah that could have likely created duplicate calls. Refactoring now 🫡

PR #103, opened 2026-02-09
chore: update thumbnail sampling strategy

monsieurBoutte:
# Overview

Replaces the simple interval-based thumbnail sampling in the moderation workflow with a smarter, duration-aware sampling strategy. When `maxSamples` is specified, timestamps are now planned using a dedicated algorithm that trims video edges, adapts cadence to duration, seeds anchor windows for diversity, and snaps to frame boundaries.

# What was changed

- **`src/lib/sampling-plan.ts`** (new) — Core sampling algorithm. `planSamplingTimestamps` computes bounded, frame-quantized candidate timestamps by trimming start/end, distributing a base grid at a duration-adaptive cadence, and sprinkling extra samples around configurable anchor percents (20%, 50%, 80%) to catch likely interesting frames.

- **`src/lib/mux-assets.ts`** — Added `getVideoTrackDurationSecondsFromAsset` to extract duration from the video track specifically, rather than relying solely on the top-level asset duration.

- **`src/workflows/moderation.ts`** — Three changes:
  1. Duration resolution now prefers video-track duration over asset-level duration.
  2. Thumbnail generation branches on `maxSamples`: when set, uses the new `planSamplingTimestamps` + a new `getThumbnailUrlsFromTimestamps` helper (which handles signing); when unset, falls back to the existing interval-based `getThumbnailUrls`.
  3. Usage metadata reports the resolved duration instead of the raw asset duration.

- **`tests/unit/sampling-plan.test.ts`** (new) — Unit tests covering: empty result when trimming removes all usable span, bounded/sorted/unique output guarantees, frame-boundary quantization, and 2-decimal-place precision on frame-rounded values.

# Suggested review order

1. `src/lib/sampling-plan.ts` — Understand the algorithm first; everything else plugs into it.
2. `tests/unit/sampling-plan.test.ts` — Verify the algorithm's contract and edge cases.
3. `src/lib/mux-assets.ts` — Small addition; note the defensive typing on the video track.
4. `src/workflows/moderation.ts` — See how the pieces integrate; pay attention to the `maxSamples` branching logic and the new `getThumbnailUrlsFromTimestamps` helper.

# Key design decisions

- **Two code paths, not one**: The existing interval-based path (`getThumbnailUrls`) is preserved when `maxSamples` is not provided, keeping backward compatibility. The new path only activates when `maxSamples` is explicitly set.
- **Video-track duration preferred**: Asset-level duration can include audio-only tails; using the video track's duration gives more accurate sampling bounds for thumbnail extraction.
- **Frame-snapping with 2-decimal precision**: All timestamps are rounded to the nearest frame boundary and then to 2 decimal places, avoiding floating-point artifacts (e.g. with non-integer fps like 29.97) and producing stable, cache-friendly thumbnail URLs.

### DIFF
--- a/src/lib/mux-assets.ts
+++ b/src/lib/mux-assets.ts
@@ -90,3 +90,9 @@ export function getAssetDurationSecondsFromAsset(asset: MuxAsset): number | unde
   const duration = asset.duration;
   return typeof duration === "number" && Number.isFinite(duration) ? duration : undefined;
 }
+
+export function getVideoTrackDurationSecondsFromAsset(asset: MuxAsset): number | undefined {
+  const videoTrack = asset.tracks?.find(track => track.type === "video");
+  const duration = (videoTrack as { duration?: unknown } | undefined)?.duration;
+  return typeof duration === "number" && Number.isFinite(duration) ? duration : undefined;
+}
--- a/src/lib/mux-assets.ts
+++ b/src/lib/mux-assets.ts
@@ -96,3 +96,11 @@ export function getVideoTrackDurationSecondsFromAsset(asset: MuxAsset): number |
   const duration = (videoTrack as { duration?: unknown } | undefined)?.duration;
   return typeof duration === "number" && Number.isFinite(duration) ? duration : undefined;
 }
+
+export function getVideoTrackMaxFrameRateFromAsset(asset: MuxAsset): number | undefined {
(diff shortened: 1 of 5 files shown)


philcluff (approved):
Looks good! Let's ship it and see how it holds up, and we can always tune from there!

PR #105, opened 2026-02-10
Add Authorization Token support

daniel-hayes:


src/env.ts
```diff
@@ -70,15 +70,7 @@ const EnvSchema = z.object({
     "Full URL for posting Evalite results (e.g., https://example.com/api/evalite-results).",
     "Evalite results endpoint",
   ),
-}).refine(
```
daniel-hayes, line 73: since we no longer need to rely entirely on env variables, we don't need to be strict here.

PR #109, opened 2026-02-12
feat: add hotspot primitive

walkerfrankenberg:
Fixes https://mux1.atlassian.net/browse/AIT-25

This PR adds a new hotspots primitive that provides access to the engagement API, allowing you to identify the most engaging time ranges in a video.

Adds three public functions for fetching hotspots by different identifiers:

getHotspotsForAsset(assetId, options) - Fetch by Mux asset ID
getHotspotsForVideo(videoId, options) - Fetch by Mux video ID
getHotspotsForPlaybackId(playbackId, options) - Fetch by playback ID

Configuration options (Data API options):

limit - Maximum number of hotspots (default: 5)
orderDirection - Sort order: 'asc' or 'desc' (default: 'desc')
orderBy - Sort field (currently only 'score')
timeframe - Time window for data (default: '[24:hours]')
credentials - Optional workflow credentials

### DIFF
--- a/src/primitives/hotspots.ts
+++ b/src/primitives/hotspots.ts
@@ -1,0 +1,152 @@
+import { getMuxClientFromEnv } from "@mux/ai/lib/client-factory";
+import type { WorkflowCredentialsInput } from "@mux/ai/types";
+
+export interface Hotspot {
+  /** Inclusive start time in milliseconds */
+  startMs: number;
+  /** Exclusive end time in milliseconds */
+  endMs: number;
+  /** Hotspot score using distribution-based normalization (0-1) */
+  score: number;
+}
+
+export interface HotspotOptions {
+  /** Maximum number of hotspots to return (default: 5) */
+  limit?: number;
+  /** Sort order: 'asc' or 'desc' (default: 'desc') */
+  orderDirection?: "asc" | "desc";
+  /** Order by field (default: 'score') */
+  orderBy?: "score";
+  /** Time window for results, e.g., ['7:days'] (default: ['7:days']) */
+  timeframe?: string;
+  /** Optional workflow credentials */
+  credentials?: WorkflowCredentialsInput;
+}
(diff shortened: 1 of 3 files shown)


src/primitives/hotspots.ts
```diff
@@ -0,0 +1,152 @@
...
+  identifierType: "assets" | "videos" | "playback-ids",
+  id: string,
+  options: HotspotOptions,
+): Promise<HotspotResponse> {
+  "use step";
+  const {
+    limit = 5,
+    orderDirection = "desc",
+    orderBy = "score",
+    timeframe = "[24:hours]",
```
monsieurBoutte, line 133: 💭 Is there an upper bound that we should document here?

src/primitives/hotspots.ts
```diff
@@ -0,0 +1,152 @@
...
+    credentials,
+  } = options;
+
+  const muxClient = await getMuxClientFromEnv(credentials);
+  const mux = await muxClient.createClient();
```
monsieurBoutte, line 138: 💭 This is probably fine for now, but I'm wondering if the client should be passed in as an argument so that the workflow can stipulate how the client gets Instantiated.

monsieurBoutte (approved):
this looks great 👍

src/primitives/hotspots.ts
walkerfrankenberg: There is an upper bound which is defined by the Data API / TTL (100 days). Think we'd likely want to document it somewhere, let's chat about docs and where makes sense in standup

src/primitives/hotspots.ts
walkerfrankenberg: Yep, agreed. When we start using these in workflows, I'll update this (and the heatmap primitive) to do that. For now, instantiating it in here will make it easier to test 👍

PR #122, opened 2026-02-18
fix: Fix incorrect categories in Hive

philcluff:
We had some categories in our list for Hive that were either wrong or didn't exist. 

I checked these by hand against our API responses.

### DIFF
--- a/src/workflows/moderation.ts
+++ b/src/workflows/moderation.ts
@@ -112,11 +112,10 @@ const DEFAULT_PROVIDER = "openai";
 const HIVE_ENDPOINT = "https://api.thehive.ai/api/v2/task/sync";
 const HIVE_SEXUAL_CATEGORIES = [
   "general_nsfw",
-  "general_suggestive",
   "yes_sexual_activity",
-  "sex_toys",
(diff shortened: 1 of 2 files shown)


src/workflows/moderation.ts
```diff
@@ -128,10 +127,8 @@ const HIVE_VIOLENCE_CATEGORIES = [
   "hanging",
   "noose",
   "human_corpse",
-  "emaciated_body",
-  "self_harm",
-  "animal_abuse",
-  "fights",
+  "yes_emaciated_body",
+  "yes_self_harm",
   "garm_death_injury_or_military_conflict",
```
walkerfrankenberg, line 132: is this supposed to be `garm`

walkerfrankenberg (commented):
lgtm

walkerfrankenberg (approved):

src/workflows/moderation.ts
philcluff: Yeh, GARM is (was?) a brand safety framework for advertisers - Hive has categories for their definitions

PR #123, opened 2026-02-18
chore: add dynamic model selection for evals

monsieurBoutte:
# Overview

This branch makes eval model coverage configurable at runtime instead of always running all configured models. It adds env/CLI/workflow controls so eval runs can target provider defaults, all models, or an explicit model list.

Additionally this change ensures when we run evals for all _supported_ models when publishing results -- post merge to main

# What was changed

- Added runtime eval model resolution logic in `src/lib/providers.ts`:
  - new `resolveEvalModelConfigs(...)`
  - new `resolveEvalModelConfigsFromEnv(...)`
  - `EVAL_MODEL_CONFIGS` now resolves from env
- Added eval env vars in `src/env.ts`:
  - `MUX_AI_EVAL_MODEL_SET` (`default` | `all`)
  - `MUX_AI_EVAL_MODELS` (comma-separated `provider:model` pairs, overrides model set)
- Updated `scripts/export-evalite-results.ts`:
  - added CLI options `--model-set` and `--models`
  - sets env vars before running Evalite
  - logs available/resolved model configs for visibility
- Updated CI workflows:
  - `.github/workflows/eval-tests.yml` adds `workflow_dispatch` model-set input and passes `MUX_AI_EVAL_MODEL_SET`
  - `.github/workflows/publish-evalite-results.yml` adds matching dispatch input and env wiring
- Updated scripts/docs:
  - `package.json` eval post-results scripts now run with `--model-set all`
  - `docs/EVALS.md` documents default/all/explicit model selection behavior and commands

# Suggested review order

1. `src/lib/providers.ts` (core selection + validation behavior)
2. `scripts/export-evalite-results.ts` (CLI surface and env precedence)
3. `.github/workflows/eval-tests.yml` and `.github/workflows/publish-evalite-results.yml` (CI/manual trigger wiring)
4. `src/env.ts` (new env contract)
5. `docs/EVALS.md` and `package.json` (usage and operational defaults)

### DIFF
--- a/scripts/export-evalite-results.ts
+++ b/scripts/export-evalite-results.ts
@@ -1,12 +1,11 @@
-/* eslint-disable node/no-process-env */
-
 import { Command } from "commander";
 import { runEvalite } from "evalite/runner";
 
 import { LANGUAGE_MODELS, resolveEvalModelConfigsFromEnv } from "../src/lib/providers";
+import type { EvalModelSelection } from "../src/lib/providers";
 
 interface Options {
-  modelSet: "default" | "all";
+  modelSet: EvalModelSelection;
   models?: string;
 }
 
--- a/scripts/export-evalite-results.ts
+++ b/scripts/export-evalite-results.ts
@@ -1,7 +1,41 @@
+/* eslint-disable node/no-process-env */
+
+import { Command } from "commander";
 import { runEvalite } from "evalite/runner";
 
-async function runEvals() {
+import { LANGUAGE_MODELS, resolveEvalModelConfigsFromEnv } from "../src/lib/providers";
+
+interface Options {
+  modelSet: "default" | "all";
+  models?: string;
+}
+
+function formatResolvedModelConfigs() {
+  return resolveEvalModelConfigsFromEnv()
+    .map(({ provider, modelId }) => `${provider}:${modelId}`)
+    .join(", ");
+}
+
+async function runEvals(options: Options) {
   try {
+    process.env.MUX_AI_EVAL_MODEL_SET = options.modelSet;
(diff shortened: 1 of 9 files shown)


scripts/export-evalite-results.ts
```diff
@@ -1,7 +1,41 @@
+/* eslint-disable node/no-process-env */
```
walkerfrankenberg: should we remove this rule rather than disabling it for the entire file? Notice that we use `process.env` throughout this repo

scripts/export-evalite-results.ts
```diff
@@ -1,7 +1,41 @@
...
+
+import { Command } from "commander";
 import { runEvalite } from "evalite/runner";
 
-async function runEvals() {
+import { LANGUAGE_MODELS, resolveEvalModelConfigsFromEnv } from "../src/lib/providers";
+
+interface Options {
+  modelSet: "default" | "all";
```
walkerfrankenberg: nit: should we use `EvalModelSelection` type from below

walkerfrankenberg (approved):
two small nits but not blockers, lgtm

scripts/export-evalite-results.ts
monsieurBoutte: in most places we actually want to enforce it. Perhaps another approach is to update at the config level which directories it's allowed in

scripts/export-evalite-results.ts
monsieurBoutte: yup, good spot

scripts/export-evalite-results.ts
monsieurBoutte: committed that change

PR #128, opened 2026-02-23
feat: Simple control over title and description length.

philcluff:
## Summary

Adds three optional params to `SummarizationOptions` — `titleLength`, `descriptionLength`, `tagCount` — so consumers can control output constraints without resorting to full prompt overrides.

## Changes

- **`titleLength`** (chars): Replaces the default "under 10 words" title brevity hint
- **`descriptionLength`** (chars): Replaces the default "2-4 sentences" description constraint
- **`tagCount`**: Controls both the prompt keyword limit and the `normalizeKeywords` cap

The static prompt builders are now factory functions (`createSummarizationBuilder`, `createAudioOnlyBuilder`) that accept a `PromptConstraints` object, so the template content is parameterised at construction rather than mutated after the fact.

All params are optional. Defaults preserve existing behaviour. User `promptOverrides` still take precedence.

The basic example (`examples/summarization/basic-example.ts`) has been updated with `--title-length`, `--description-length`, and `--tag-count` CLI flags for manual testing.

## Test plan

- [x] `npm run build` — no type errors
- [x] `npm run lint` — clean
- [x] All unit tests pass unchanged

### DIFF
--- a/examples/summarization/basic-example.ts
+++ b/examples/summarization/basic-example.ts
@@ -21,11 +21,17 @@ program
   .option("-m, --model <model>", "Model name (overrides default for provider)")
   .option("-t, --tone <tone>", "Tone for summary (neutral, playful, professional)", "neutral")
   .option("--no-transcript", "Exclude transcript from analysis")
+  .option("--title-length <chars>", "Desired title length in characters", parseInt)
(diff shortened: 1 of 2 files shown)


monsieurBoutte (approved):
this looks good to me 👍

PR #129, opened 2026-02-23
chore: tune system prompts to better utilize caching

walkerfrankenberg:
# Overview

Improves prompt caching hit rates for two workflows by ensuring stable prompt content is always in the system prompt and dynamic content is always appended last or in the user prompt. Based off of [OpenAI's suggestions for prompt caching](https://developers.openai.com/cookbook/examples/prompt_caching_201)

# What was changed

**[src/workflows/ask-questions.ts](src/workflows/ask-questions.ts)**
- Removed `buildSystemPrompt()`, which concatenated the dynamic `allowedAnswers` list onto the end of `SYSTEM_PROMPT`. Every distinct set of allowed answers produced a unique system prompt string, busting the cache on every variation.
- `SYSTEM_PROMPT` is now used directly and is identical across all requests.
- `<response_options>` (the allowed answers list) is appended to the user prompt instead, where dynamic content belongs.

**[src/workflows/translate-captions.ts](src/workflows/translate-captions.ts)**
- The entire prompt was a single user message with `fromLanguageCode`, `toLanguageCode`, and `vttContent` all inlined, leaving zero stable cacheable prefix.
- Extracted the static instructions into a `SYSTEM_PROMPT` constant shared across all translation requests regardless of language pair.
- User message is now just `"Translate from {from} to {to}:\n\n{vttContent}"`.


src/workflows/translate-captions.ts
```diff
@@ -123,9 +133,13 @@ async function translateVttWithAI({
...
     output: Output.object({ schema: translationSchema }),
     messages: [
+      {
+        role: "system",
+        content: SYSTEM_PROMPT,
+      },
       {
         role: "user",
-        content: `Translate the following VTT subtitle file from ${fromLanguageCode} to ${toLanguageCode}. Preserve all timestamps and VTT formatting exactly as they appear. Return JSON with a single key "translation" containing the translated VTT.\n\n${vttContent}`,
+        content: `Translate from ${fromLanguageCode} to ${toLanguageCode}:\n\n${vttContent}`,
```
walkerfrankenberg, line 139: this change won't actually help any prompt caching since the system prompt is too short, but this is more consistent with our other prompts' structures

src/workflows/translate-captions.ts
```diff
@@ -85,6 +85,16 @@ export const translationSchema = z.object({
...
 
+// ─────────────────────────────────────────────────────────────────────────────
+// Prompts
+// ─────────────────────────────────────────────────────────────────────────────
+
+const SYSTEM_PROMPT =
+  "You are a subtitle translation expert. " +
+  "Translate VTT subtitle files to the target language specified by the user. " +
+  "Preserve all timestamps and VTT formatting exactly as they appear. " +
+  "Return JSON with a single key \"translation\" containing the translated VTT content.";
```
monsieurBoutte: 💭 should we be making use of `dedent` here as well for consistency?

walkerfrankenberg: Tbh I split this up into multiple lines unnecessarily for readability. My understanding was that `dedent` is for multiline template literals where indentation gets baked in whereas this prompt is a single logical string split across lines only for readability. I can put it back into one line if that's preferred

src/workflows/ask-questions.ts
```diff
@@ -201,17 +192,23 @@ function buildUserPrompt(
...
 
   const format = isCleanTranscript ? "plain text" : "WebVTT";
   const transcriptSection = createTranscriptSection(transcriptText, format);
 
-  return askQuestionsPromptBuilder.buildWithContext(
+  return `${askQuestionsPromptBuilder.buildWithContext(
     { questions: questionsContent },
     [transcriptSection],
-  );
+  )}\n\n${responseOptions}`;
```
monsieurBoutte: 💭  i assume the thinking here is that answer options might be changed on subsequent request therefore if we put them lower in the prompt, we could increase the chances of cache hits -- I think that's a great idea

monsieurBoutte: should we also consider the same for questions?

I'm guessing optimal sequencing here is

- SYSTEM PROMPT
- TRANSCRIPT 
- ACTUAL QUESTION(S)
- POSSIBLE ANSWERS

walkerfrankenberg: yes that, and also removing the dynamic user-inputted content from the system prompt so that the system prompt is always static and thus will always be a cache hit

monsieurBoutte (approved):
no blocking feedback but left some thoughts for consideration

PR #140, opened 2026-03-10
feat: Return clearer timing information from moderation workflows

philcluff:
## Summary

Adds `time?: number` to `ThumbnailModerationScore` so consumers can see when in the video a moderation flag occurred without parsing it back out of the thumbnail URL (which is impossible for signed URLs).

**Breaking change:** `getThumbnailUrls` primitive now returns `Array<{ url: string; time: number }>` instead of `string[]`. Any code that treats the return values as strings (e.g. `urls[0].includes(...)`, `fetch(urls[0])`) will need updating to access the `.url` property.

## Example

```typescript
const result = await getModerationScores("asset-id", { provider: "openai" });

for (const score of result.thumbnailScores) {
  if (score.sexual > 0.8) {
    console.log(`Flagged at ${score.time}s: ${score.url}`);
  }
}
```

## Suggested review order

1. `src/primitives/thumbnails.ts` — return type change
2. `src/workflows/moderation.ts` — threading `time` through OpenAI and Hive paths
3. `tests/unit/thumbnails.test.ts` — updated assertions
4. `tests/integration/signed-playback.test.ts` — same
5. `docs/API.md`, `docs/PRIMITIVES.md` — updated type docs
6. `examples/signed-playback/signed-playback.ts` — updated example

### DIFF
--- a/src/workflows/moderation.ts
+++ b/src/workflows/moderation.ts
@@ -28,6 +28,8 @@ import type {
 /** Per-thumbnail moderation result returned from `getModerationScores`. */
 export interface ThumbnailModerationScore {
   url: string;
+  /** Time in seconds of the thumbnail within the video. Absent for transcript moderation entries. */
+  time?: number;
   sexual: number;
   violence: number;
   error: boolean;
@@ -153,6 +155,7 @@ async function processConcurrently<T>(
 
 async function moderateImageWithOpenAI(entry: {
   url: string;
+  time?: number;
   image: string;
   model: string;
   credentials?: WorkflowCredentialsInput;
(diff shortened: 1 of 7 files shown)


daniel-hayes (approved):

src/workflows/moderation.ts
```diff
@@ -28,6 +28,8 @@ import type {
 /** Per-thumbnail moderation result returned from `getModerationScores`. */
 export interface ThumbnailModerationScore {
   url: string;
+  /** Time in seconds of the thumbnail within the video. Absent for transcript moderation entries. */
+  time?: number;
```
monsieurBoutte, line 32: 💭I'm curious about whether or not this skills be declared as optional. I didn't see any optional chaining in the code work fallbacks

philcluff: Chatted a bit in Slack:
* This is optional because of audio_only support, which doesn't return a time
* However, having thumbnailScores at all in this shape doesn't make much sense for audio_only, and is probably wrong

Going to ticket having a better shape here for audio only and roll forward.

PR #155, opened 2026-04-03
fix: update engagement insights when no engagement data is present

walkerfrankenberg:
# Overview

When a video has no engagement data (e.g., no views yet), the engagement insights workflow now returns a graceful empty result instead of throwing an error. This gives callers a structured response with an explanatory summary they can surface to users.

# What was changed

- **`src/workflows/engagement-insights.ts`** — Replaced `throw new Error(...)` with an early return of `{ assetId, momentInsights: [], overallInsight: { summary: "...", trends: [] } }` when hotspots are empty.
- **`tests/integration/engagement-insights.test.ts`** — Updated the no-engagement-data test from expecting a thrown error to asserting on the early-return shape.
- **`package.json` / `package-lock.json`** — Version bump `0.13.0` → `0.13.1`.

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) is generating a summary for commit d8fe66637a1ecf80abbfac696b8cfb773af3bb71. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/src/workflows/engagement-insights.ts
+++ b/src/workflows/engagement-insights.ts
@@ -612,10 +612,16 @@ export async function generateEngagementInsights(
   const { hotspots, heatmap } = engagementResult.value;
 
   if (hotspots.length === 0) {
-    throw new Error(
-      `No engagement data available for asset ${assetId} in timeframe ${timeframe}. ` +
-      `Video may not have been viewed yet or engagement tracking is not enabled.`,
-    );
+    return {
+      assetId,
+      momentInsights: [],
+      overallInsight: {
+        summary:
+          `No engagement data available for asset ${assetId} in timeframe ${timeframe}. ` +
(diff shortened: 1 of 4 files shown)


monsieurBoutte (approved):

src/workflows/engagement-insights.ts
```diff
@@ -612,10 +612,16 @@ export async function generateEngagementInsights(
   const { hotspots, heatmap } = engagementResult.value;
```
monsieurBoutte, line 612: any additional guards needed for heatmap?

walkerfrankenberg: heatmap will always be an array and will be present, so that should be safe. If 's an empty array, then that _would_ get passed in regardless - but the only case it'd be empty is if there's no views, and if that's the case, then there will be no hotspots anyways.

PR #186, opened 2026-04-23
feat: Add case sensitivty option in static replacements in `edit-captions`

philcluff:
## Summary

Adds an optional per-entry `caseSensitive` flag to the `replacements` array in `editCaptions`. Defaults to `true`, preserving current behaviour; set to `false` to match regardless of case.

The existing case-sensitive default made sense for specific brand-name corrections (`"Mucks" → "Mux"`), but was clumsy for anything where the caption source varies in case — e.g. a filler like `gonna` that might appear as `Gonna` at the start of a cue.

## Example

```typescript
const result = await editCaptions(assetId, trackId, {
  replacements: [
    { find: "Mucks", replace: "Mux" }, // case-sensitive (default)
    { find: "gonna", replace: "going to", caseSensitive: false }, // matches "Gonna", "GONNA", etc.
  ],
});
```

Auto-censor profanity behaviour and override lists are untouched — those remain case-insensitive as before.

## Suggested review order

1. `src/workflows/edit-captions.ts` — type field + the `gi` vs `g` branch in `applyReplacements`
2. `tests/unit/edit-captions.test.ts` — four new cases covering default, explicit true/false, all-case variants, and per-entry mixing
3. `docs/API.md` and `docs/WORKFLOWS.md` — updated option docs

Note: the CLI (`examples/edit-captions/basic-example.ts`) is intentionally not extended — the `find:replace` DSL stays simple; programmatic API only.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -410,7 +410,7 @@ Edits a caption track using LLM-powered profanity censorship, static find/replac
     - `'mask'`: `shit` → `????` (question marks matching word length)
   - `alwaysCensor?: string[]` - Words to always censor regardless of LLM output
   - `neverCensor?: string[]` - Words to never censor even if the LLM flags them (takes precedence over `alwaysCensor`)
-- `replacements?: Array<{ find: string; replace: string }>` - Static find/replace pairs (optional, no LLM needed)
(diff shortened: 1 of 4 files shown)


cjpillsbury (approved):
LGTM!

PR #189, opened 2026-04-24
feat: Free-form ask questions

philcluff:
## Summary

Adds an **experimental** `freeFormReply` mode to `askQuestions`, letting individual questions opt out of the constrained-enum schema and reply with prose instead.

- Per-question opt-in (`Question.freeFormReply: true`) — a single call can mix yes/no, custom enums, and free-form answers.
- Length-capped via new `AskQuestionsOptions.maxFreeFormAnswerLength` (default **500 chars**).
- Mutually exclusive with `answerOptions` — setting both throws a validation error.
- Free-form answers go through the same `scrubFreeTextField` safety pipeline as `reasoning` — leaks flip the question to skipped and suppress the output.
- System prompts reworked to describe both `<allowed_answers>` and `<answer_format>` in a mode-neutral way, so the constrained-only path's strictness is preserved.
- CLI examples extended with a `|*` sigil and a `--free-form-max-length` flag.

## Example

```ts
const result = await askQuestions(assetId, [
  { question: "Is this video about glasses?" },                                // yes/no
  { question: "What is the primary subject?",
    answerOptions: ["glasses", "watches", "shoes", "hats"] },                  // enum
  { question: "Describe the primary subject in one sentence.",
    freeFormReply: true },                                                     // free-form
], { maxFreeFormAnswerLength: 300 });
```

CLI:

```bash
npm run example:ask-questions -- abc123 "Describe the primary subject.|*"
```

## Design notes

- **Internal model:** questions normalise into a discriminated union — ` { mode: { kind: "yesNo" | "options" | "freeForm", ... } }`. Yes/no stays implicit (no eager `["yes", "no"]` substitution); expansion only happens at the prompt and schema layer. All downstream branching is `switch (q.mode.kind)`.
- **Step boundary:** `analyzeQuestions` takes plain POJO question data plus `maxFreeFormAnswerLength` and builds the response schema inside the step. The Zod schema never crosses the workflow runtime's serialisation boundary, so there's no nullable-tuple workaround.
- **Schema tradeoff:** when any question is free-form, the answer sub-schema relaxes to `z.string().min(1).max(maxLen)` for the whole batch — Zod can't express a per-question mixed enum across a uniform array. Per-question enum enforcement for the still-constrained questions in a mixed batch falls to the existing post-processing check. Constrained-only calls retain provider-side enum enforcement.
- **Prompt rework:** the constrained-only path's "ONLY allowed response options" language is preserved but scoped per-question (i.e. only applies to questions with `<allowed_answers>`). When at least one question is free-form, an addendum carries free-form specifics (length cap, skip semantics, content-not-instructions hygiene). Default (constrained-only) prompts are semantically unchanged.

## Suggested review order

1. `src/workflows/ask-questions.ts` — types, schema builder, prompts, normalisation, post-processing.
2. `examples/ask-questions/parse-question.ts` — `|*` sigil parsing.
3. `examples/ask-questions/basic-example.ts` + `multiple-questions-example.ts` — CLI flag + display.
4. `tests/integration/ask-questions.test.ts` — new free-form + mutual-exclusivity coverage.
5. Docs (`docs/API.md`, `docs/WORKFLOWS.md`, `examples/ask-questions/README.md`).

---

> [!NOTE]
> **Medium Risk**
> Changes `askQuestions` prompt/schema generation and output post-processing to support open-ended answers, which affects a core workflow and adds a new free-text output channel despite length caps and safety scrubbing.
> 
> **Overview**
> Adds an **experimental** per-question `freeFormReply` mode to `askQuestions`, allowing a single call to mix constrained yes/no or `answerOptions` questions with open-ended prose answers.
> 
> Introduces `maxFreeFormAnswerLength` (default 500) plus new validation (mutual exclusivity with `answerOptions`), updates prompt construction to emit `<allowed_answers>` vs `<answer_format>`, relaxes the provider-side response schema when any free-form question is present, and adds post-processing to scrub and length-cap free-form answers (treating leaks/overages as skipped).
> 
> Extends the CLI examples with a `|*` free-form sigil and `--free-form-max-length`, updates docs to describe the experimental API, and adds integration tests covering free-form behavior, mixed batches, validation, and length limits.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4631c2d913a5e583d90f0ef44d193342842856ae. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/workflows/ask-questions.ts
+++ b/src/workflows/ask-questions.ts
@@ -41,22 +41,11 @@ export interface Question {
   /** Allowed answers for this question (defaults to ["yes", "no"]). */
   answerOptions?: string[];
   /**
-   * Experimental: When true, the model replies with free-form text for
-   * this question instead of yes/no or `answerOptions`. Answers are
-   * length-capped via {@link AskQuestionsOptions.maxFreeFormAnswerLength}
-   * (default 500 characters), still subject to the output-safety
-   * scrubber, and the model can still skip irrelevant questions via the
-   * normal skip mechanism.
-   *
-   * Mutually exclusive with `answerOptions`: setting both throws a
-   * validation error. Default (neither set) remains yes/no.
-   *
-   * Free-form is the more permissive mode: enum-constrained answers
-   * prevent arbitrary prose in model output, which is a potential
-   * prompt-leak exfiltration channel. Use free-form only when your use
-   * case genuinely needs open-ended answers (e.g. describing a scene,
-   * extracting quoted text) and treat the resulting answer as untrusted
-   * model output.
+   * Experimental: replies with free-form prose instead of yes/no or
+   * `answerOptions`. Length-capped via
+   * {@link AskQuestionsOptions.maxFreeFormAnswerLength} (default 500),
+   * still scrubbed for safety, still skippable. Mutually exclusive with
+   * `answerOptions`. Treat the answer as untrusted model output.
    */
   freeFormReply?: boolean;
 }
@@ -108,15 +97,9 @@ export interface AskQuestionsOptions extends MuxAIOptions {
    */
   maxAnswerOptionLength?: number;
   /**
-   * Experimental: Maximum character length for free-form answers when a
-   * question sets `freeFormReply: true`. Defaults to 500.
-   *
-   * Free-form answers bypass the enum schema, so this cap is the primary
-   * mechanical limit on how much content the model can emit per answer.
-   * Keep it low to hold answers tight and bound any potential
-   * exfiltration channel; raise only if your use case genuinely needs
-   * longer prose. Answers still pass through the output-safety scrubber
-   * regardless of this value.
+   * Experimental: max character length for free-form answers when
+   * `freeFormReply: true`. Defaults to 500. Free-form bypasses the enum
+   * schema, so this cap is the primary limit on output length.
    */
   maxFreeFormAnswerLength?: number;
 }
@@ -414,15 +397,7 @@ const AUDIO_ONLY_SYSTEM_PROMPT = promptDedent`
     - Be specific and evidence-based
   </language_guidelines>`;
 
-/**
- * Additional details for free-form (<answer_format>) questions, appended
- * to the system prompt when at least one question uses free-form mode.
- *
- * The base system prompt already describes both response formats in a
- * mode-neutral way; this block only carries free-form-specific rules
- * (length cap, skip-via-sentinel semantics, no-instructions hygiene) so
- * the default path is unaffected when no question is free-form.
- */
+// Appended to the system prompt only when free-form mode is in use.
 function freeFormAddendum(maxFreeFormAnswerLength: number): string {
   return promptDedent`
     <free_form_answers>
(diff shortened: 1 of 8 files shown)


src/workflows/ask-questions.ts
```diff
@@ -376,40 +421,98 @@ const AUDIO_ONLY_SYSTEM_PROMPT = promptDedent`
...
+ *
+ * The base system prompt already describes both response formats in a
+ * mode-neutral way; this block only carries free-form-specific rules
+ * (length cap, skip-via-sentinel semantics, no-instructions hygiene) so
+ * the default path is unaffected when no question is free-form.
+ */
+function freeFormAddendum(maxFreeFormAnswerLength: number): string {
+  return promptDedent`
+    <free_form_answers>
+      EXPERIMENTAL. Additional rules for questions with <answer_format>free-form text ...</answer_format>:
```
philcluff: I don't think `EXPERIMENTAL` belongs in a prompt.

src/workflows/ask-questions.ts
```diff
@@ -202,15 +243,16 @@ const SYSTEM_PROMPT = promptDedent`
   <task>
     For each question provided, you must:
     1. Analyze the storyboard frames and transcript (if provided)
-    2. Answer with ONLY the allowed response options listed for that question - no other values are acceptable
+    2. Answer each question according to its specified response format: for questions with <allowed_answers>, select ONLY from the listed values; for questions with <answer_format>, follow the format specification exactly
```
philcluff, line 222: I dislike that this mixes the two modes of operation into a single system prompt, but I also don't see an obvious better solution here right now, as we allow many prompts in a single call.

src/workflows/ask-questions.ts
```diff
@@ -460,7 +563,17 @@ interface AnalyzeQuestionsInput {
   modelId: string;
   userPrompt: string;
   systemPrompt: string;
-  allowedAnswers: [string, ...string[]];
+  /**
```
philcluff: I don't love this honestly, especially the serialisation caveats. Maybe there's a better structure here?

src/workflows/ask-questions.ts
```diff
@@ -674,12 +814,33 @@ export async function askQuestions(
...
-    return { ...q, answerOptions: options as [string, ...string[]] };
+    return { question: q.question, answerOptions: options as [string, ...string[]] };
   });
 
-  const allowedAnswers = Array.from(
-    new Set(normalizedQuestions.flatMap(q => q.answerOptions)),
-  ) as [string, ...string[]];
+  const hasFreeForm = normalizedQuestions.some(q => q.freeFormReply === true);
+
+  // When any question is free-form the answer sub-schema becomes a
```
philcluff: This is all a hot mess, is there not a more elegant solution to this? We're not locked into API shape here.

philcluff (commented):
The typing and optionality management here isn't great.

src/workflows/ask-questions.ts
cjpillsbury: we could parameterize this as well, but this is maybe fine for now. Also, we should probably normalize to unordered/bullet lists instead of ordered/numbered unless we think it's important for LLM context.

cjpillsbury (approved):
Looks like cursor has a legit callout for ask-questions but pre-approving because otherwise looks good enough for this iteration.

PR #193, opened 2026-04-29
chore: Bump to latest stable Vercel workflow

philcluff:
## Summary

Bumps `workflow` from `4.2.0-beta.64` → `^4.2.4`. The version bump alone broke our Nitro-based integration tests with `Package subpath './lib/X' is not defined by "exports"` errors at workflow runtime. This PR fixes that and aligns the whole `src/` tree to a single import style.

## Root cause

[`vercel/workflow#1613`](https://github.com/vercel/workflow/pull/1613) (shipped in `4.2.0-beta.78`) changed how the step bundler handles path-aliased imports of project-local helpers. Before it, esbuild's tsconfig-paths fallback **inlined** helpers (e.g. `client-factory.ts`) into the step bundle, so all transitive `@mux/ai/lib/*` self-references resolved at bundle time. After it, helpers get **externalized** as relative paths and loaded as source by Node's strict ESM resolver at runtime — and Node doesn't know about tsconfig `paths` or Nitro/Vite aliases.

Result: once `src/lib/client-factory.ts` was loaded as source, its `import ... from "@mux/ai/lib/providers"` hit Node's resolver, fell through to the package's `exports` map, didn't find `./lib/providers`, and threw.

This is a sibling of [`vercel/workflow#1654`](https://github.com/vercel/workflow/issues/1654) (Node builtins mangled into relative paths). That one was fixed by [`vercel/workflow#1644`](https://github.com/vercel/workflow/pull/1644) — its `!esbuildResult.external` guard catches builtins, but doesn't help genuinely project-local helpers like ours. The contract isn't documented anywhere I could find.

## What changed

Across the whole `src/` tree:

- All `@mux/ai/*` self-references → relative imports
- All relative imports → explicit `.ts` extensions
- `tsconfig.json` gets `"allowImportingTsExtensions": true` (only enforced by `tsc --noEmit`; tsup/esbuild handles `.ts` natively, so the published bundle is unaffected)
- Stripped the now-dead `@mux/ai/types` and `@mux/ai/*` entries from the tsconfig `paths` block; kept the three public entries (`@mux/ai`, `@mux/ai/workflows`, `@mux/ai/primitives`) used by tests/examples/intellisense
- Added two scoped ESLint rules to enforce the style going forward — `no-restricted-imports` blocks `@mux/ai/*` inside `src/`, and `no-restricted-syntax` requires `.ts` on relative imports/exports. Built-ins only, no new deps. Both rules carry an explanatory message so future-us doesn't wonder why.


daniel-hayes (approved):

decepulis (approved):

walkerfrankenberg (approved):

cjpillsbury (approved):
LGTM!

PR #224, opened 2026-07-31
fix: surface content policy errors from summaries

monsieurBoutte:
# Overview

Surface content-policy failures from summary generation as explicit, non-retryable errors instead of the generic `Invalid JSON response`. Public messaging stays provider-agnostic and does not expose raw provider responses or submitted content.

# What was changed

- Normalize prompt blocks, candidate-side policy blocks, and AI SDK `content-filter` finishes into `content_policy_error`.
- Preserve branded errors across Workflow step serialization so downstream consumers receive the reason or category and repeated retries are skipped.
- Apply the handling to video and audio-only summaries, with focused coverage for blocked and unrelated failure paths.

# Suggested review order

1. `src/lib/content-policy-error.ts` — classification and safe public messaging.
2. `src/lib/mux-ai-error.ts` — error type and Workflow-boundary preservation.
3. `src/workflows/summarization.ts` — summary integration.
4. `tests/unit/content-policy-error.test.ts` — expected behavior and regressions.

---

> [!NOTE]
> **Medium Risk**
> Changes customer-visible error classification and retry behavior for summarization LLM calls; scope is limited to error handling, not core data or auth paths.
> 
> **Overview**
> Summary generation now turns provider **content-policy blocks** into explicit, **non-retryable** `content_policy_error` responses instead of opaque failures like `Invalid JSON response`.
> 
> A new **`content-policy-error`** helper parses AI SDK `APICallError` bodies (prompt `blockReason`, candidate `finishReason`, blocked safety categories) and `NoObjectGeneratedError` with `content-filter`, then raises a branded **`MuxAiError`** with provider-agnostic messaging (reason/category only, no raw responses).
> 
> **`MuxAiError`** gains the `content_policy_error` type and a **`MuxAiError.is()`** duck-type check so **`wrapError`** still rethrows serialized workflow-step errors without wrapping them as internal failures. Both video storyboard and audio-only summary **`generateText`** paths use **`withContentPolicyErrorHandling`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 96ef82de89e213c26c086c43d09b3f60c13aae66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/lib/content-policy-error.ts
+++ b/src/lib/content-policy-error.ts
@@ -28,6 +28,16 @@ const ContentPolicyResponseSchema = z.object({
   })).max(8).nullish(),
 });
 
+const ContentPolicyUsageResponseSchema = z.object({
+  usageMetadata: z.object({
+    promptTokenCount: z.number().finite().nonnegative().nullish(),
+    candidatesTokenCount: z.number().finite().nonnegative().nullish(),
+    totalTokenCount: z.number().finite().nonnegative().nullish(),
+    cachedContentTokenCount: z.number().finite().nonnegative().nullish(),
+    thoughtsTokenCount: z.number().finite().nonnegative().nullish(),
+  }).nullish(),
+});
+
 const CONTENT_POLICY_REASONS = new Set([
   "BLOCKLIST",
   "ESCALATION",
@@ -47,6 +57,13 @@ export interface ContentPolicyBlock {
   category?: string;
 }
 
+interface GeneratedOutput<T> {
+  finishReason: string;
+  rawFinishReason?: string;
+  output: T;
+  usage?: unknown;
+}
+
 export async function withContentPolicyErrorHandling<T>(operation: () => Promise<T>): Promise<T> {
   try {
     return await operation();
@@ -61,6 +78,29 @@ export function rethrowContentPolicyError(error: unknown): never {
     throw error;
   }
 
+  throwContentPolicyError(block, getContentPolicyTokenUsage(error));
+}
+
+/**
+ * Reads structured output only after checking for a successful provider
+ * response that stopped because of a content policy. AI SDK intentionally
+ * leaves output unresolved for non-`stop` finishes, so accessing `output`
+ * first would throw a metadata-free `NoOutputGeneratedError`.
+ */
+export function getGeneratedOutputWithContentPolicyHandling<T>(response: GeneratedOutput<T>): T {
+  if (response.finishReason === "content-filter") {
+    const rawReason = PolicyTokenSchema.safeParse(response.rawFinishReason);
+    const reason = rawReason.success && isContentPolicyReason(rawReason.data) ?
+      rawReason.data :
+      "CONTENT_FILTER";
+
+    throwContentPolicyError({ reason }, getErrorTokenUsage(response));
+  }
+
+  return response.output;
+}
+
+function throwContentPolicyError(block: ContentPolicyBlock, usage?: TokenUsage): never {
   const category = block.category ? `; category: ${block.category}` : "";
   const contentPolicyError = new MuxAiError(
     `The supplied content was blocked by a content policy (reason: ${block.reason}${category}).`,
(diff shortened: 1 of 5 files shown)


src/workflows/summarization.ts
```diff
@@ -570,7 +571,7 @@ async function analyzeStoryboard(
         ],
       },
     ],
-  });
+  }));
```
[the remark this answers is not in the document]
monsieurBoutte: Addressed in the pending follow-up changes. `withRetry` now retains and rethrows the original thrown value, using a normalized `Error` only for retry decisions and logging. That preserves the serialized `__robots_error` brand, `content_policy_error`, and `retryable: false` through to `wrapError` on the URL-based Workflow path.

I added a regression test for that exact path and also preserved token usage when normalizing the policy error. Local lint, typecheck, all 400 unit tests, and build pass; a fresh read-only Cursor Bugbot pass against the updated diff reports `BUGBOT GREEN`.

src/workflows/summarization.ts
```diff
@@ -570,7 +571,7 @@ async function analyzeStoryboard(
...
 
   if (!response.output) {
     throw new Error("Summarization output missing");
```
philcluff, line 582: [Claude]  Output-side Gemini content blocks escape `withContentPolicyErrorHandling`: `generateText` resolves successfully when `finishReason` is `"content-filter"`, and the error is thrown by the `response.output` getter outside the wrapper.

src/lib/content-policy-error.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  const category = block.category ? `; category: ${block.category}` : "";
+  const contentPolicyError = new MuxAiError(
+    `The supplied content was blocked by a content policy (reason: ${block.reason}${category}).`,
+    {
+      type: "content_policy_error",
+      retryable: false,
+    },
+  );
+  const usage = getErrorTokenUsage(error);
```
philcluff: [Claude] `getErrorTokenUsage` reads only `error.usage`, which `APICallError` never carries, while the billed `usageMetadata` sits unread in the same `responseBody` the code already parses — so usage is always dropped on the only reachable detection path.

src/workflows/summarization.ts
```diff
@@ -550,7 +551,7 @@ async function analyzeStoryboard(
   const model = await createLanguageModelFromConfig(provider, modelId, credentials);
   const schema = buildSummarySchema(descriptionLength);
 
-  const response = await generateText({
+  const response = await withContentPolicyErrorHandling(() => generateText({
```
philcluff, line 557: [Claude] Content-policy handling is applied by wrapping only 2 of 9 `generateText` call sites, so the same blocked asset yields `content_policy_error` from summarization but an opaque `processing_error` from every other workflow.

philcluff (commented):
Looks widely fine, but Claude caught a couple of bits that look legit to me - we've probably blocked what we _know_ fails now, but are probably liable to have similar issues in other areas that we aren't covering.

src/workflows/summarization.ts
monsieurBoutte: Good call—the broader inconsistency is real. I’d like to keep this PR scoped to fully fixing summarization, which is where we observed the failure, and track shared content-policy handling across the remaining `generateText` workflows as follow-up work rather than expand this PR’s risk and test surface.

src/workflows/summarization.ts
monsieurBoutte: Addressed in 9be23d0. We now inspect response.finishReason before reading the lazy output getter, convert content-filter responses into non-retryable content_policy_error errors, and preserve the response usage. A regression test verifies the getter is never touched on this path.

src/lib/content-policy-error.ts
monsieurBoutte: Addressed in 9be23d0. For APICallError policy blocks, we now parse and validate Gemini usageMetadata from responseBody, normalize input, output, total, reasoning, and cached token counts, and attach them to the content_policy_error. Added coverage for that exact response-body path.

philcluff (approved):

PR #233, opened 2026-08-11
feat: Support `never_translate` phrases

philcluff:
## Summary

Adds a `neverTranslate` option to `translateCaptions` for keeping brand names, product names, and other proper nouns verbatim in translated captions.

- Terms are validated at the workflow boundary (max 100 terms, 100 chars each, trimmed, case-insensitively deduped, `<`/`>` rejected) and treated as partly trusted input per `docs/SECURITY.md`.
- Enforcement is prompt-based: a `<never_translate>` section is injected into the user message on both the whole-VTT and cue-chunked translation paths. Terms are injected unescaped so the spelling the model is told to preserve is identical to the spelling verification counts; breakout is prevented by the angle-bracket rejection instead of XML escaping.
- After translation, each term's source occurrence count (case-insensitive, cue text only) is compared against verbatim occurrences in the output. Any shortfall sets `result.neverTranslateTermsPreserved` to `false` — never auto-repaired, since we can't know what the model rendered a term as.

Known limitation: per the WebVTT spec `&` arrives entity-escaped in cue text (`AT&amp;T`), so `&`-containing terms typically count zero source occurrences and verification silently skips them. Fixing that properly means entity-decoding in `extractTextFromVTT`, out of scope here.

## Example

```ts
const result = await translateCaptions(assetId, trackId, "es", {
  provider: "openai",
  neverTranslate: ["Mux", "AT&T"],
});

if (result.neverTranslateTermsPreserved === false) {
  // At least one term did not survive translation verbatim
}
```

## Suggested review order

1. `src/workflows/translate-captions.ts` — types, `validateNeverTranslateTerms`, `buildNeverTranslateSection`, `verifyNeverTranslateTerms`, then the plumbing through the step functions and the audit in `translateCaptionsInternal`
2. `tests/unit/translate-captions.test.ts` — validation and verification coverage
3. `tests/eval/translate-captions.eval.ts` — `never-translate-compliance` scorer
4. `docs/API.md`, `docs/WORKFLOWS.md`, `docs/SECURITY.md`

---

> [!NOTE]
> **Medium Risk**
> New partly-trusted strings reach the model prompt; validation mitigates injection, but compliance is prompt-based and not guaranteed.
> 
> **Overview**
> Adds **`neverTranslate`** to `translateCaptions` so callers can keep brand names and proper nouns verbatim in translated captions.
> 
> Terms are **validated** at the workflow boundary (caps, trim/dedupe, reject `<`/`>` and NFKC-altering characters) and documented as partly trusted prompt input. Both whole-VTT and cue-chunked paths inject a `<never_translate>` block into the user message and extend system prompts to require verbatim preservation.
> 
> After translation, **`verifyNeverTranslateTerms`** compares case-insensitive source counts in cue text to verbatim counts in the output; the result exposes **`neverTranslateTermsPreserved`** (audit only—no auto-repair). Docs, the CLI example (`--never-translate`), unit tests, and an eval scorer **`never-translate-compliance`** are updated accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4e23db119b1c7ccc058cda3981f4c861036c8d3e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/workflows/translate-captions.ts
+++ b/src/workflows/translate-captions.ts
@@ -101,11 +101,9 @@ export interface TranslationResult {
    */
   safety?: SafetyReport;
   /**
-   * Present when `neverTranslate` terms were supplied. Reports whether
-   * each term survived translation verbatim. Enforcement is prompt-based,
-   * so compliance is verified rather than guaranteed; `violations` lists
-   * terms that appear fewer times in the translated output than in the
-   * source. No automatic repair is attempted.
+   * Present when `neverTranslate` terms were supplied. Enforcement is
+   * prompt-based, so compliance is verified rather than guaranteed;
+   * violations are reported, never repaired.
    */
   neverTranslate?: NeverTranslateReport;
 }
@@ -121,9 +119,7 @@ export interface NeverTranslateViolation {
 
 /** Compliance report for the `neverTranslate` option. */
 export interface NeverTranslateReport {
-  /** The validated terms that were sent with the translation request. */
   terms: string[];
-  /** Terms whose verbatim occurrence count dropped during translation. */
   violations: NeverTranslateViolation[];
 }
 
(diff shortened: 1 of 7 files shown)


src/workflows/translate-captions.ts
```diff
@@ -220,6 +239,90 @@ const CUE_TRANSLATION_SYSTEM_PROMPT = promptDedent`
...
+/**
+ * Compares extracted cue text only: source occurrences count
+ * case-insensitively, translated occurrences must be verbatim.
+ */
+export function verifyNeverTranslateTerms(
+  terms: string[],
+  sourceVtt: string,
+  translatedVtt: string,
+): boolean {
+  const sourceText = extractTextFromVTT(sourceVtt).toLowerCase();
```
monsieurBoutte: Could we compare against cue text without normalization? extractTextFromVTT rewrites Unicode and whitespace, so some missing terms can get an expected count of zero and falsely pass.

philcluff: Good spot that the normalisation can hide misses. Raw comparison has the opposite problem though: cue text is NFKC-normalised before the model sees it (chunked path), so it can't preserve a character the sanitiser rewrites.

I've updated, validation now rejects terms that `sanitizeUntrustedText` would change - accepted terms are then identical in the prompt, the payload, and the verification text.

monsieurBoutte (approved):
lgtm

PR #174, opened 2026-04-16
refactor: consolidate shared prompt fragments to prevent drift

walkerfrankenberg:
## Summary
- Extracted 7 duplicated prompt fragments from 4 workflow files into a new shared module at `src/lib/prompt-fragments.ts`
- Fragments include: metadata boundary warning, no-fabrication constraint, structured-data constraint, storyboard frame instructions, tone guidance, confidence scoring rubric, and language guidelines (video/audio variants)
- Eliminates ~29 instances of copy-pasted prompt text across `summarization.ts`, `ask-questions.ts`, `engagement-insights.ts`, and `burned-in-captions.ts`

## Motivation
These prompt fragments were independently maintained across workflow files, creating risk of prompt drift — where updates to one copy diverge from the others over time. Now there is a single source of truth for each shared instruction.

## Test plan
- [x] TypeScript type-check passes (`tsc --noEmit`)
- [x] ESLint passes with zero errors/warnings
- [x] All 262 unit tests pass (`vitest run tests/unit/`)
- [ ] Integration tests pass in CI

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Primarily a refactor, but it changes how system prompts are rendered (shared fragments + `alignValues` dedent), which can subtly affect LLM behavior and structured-output compliance across multiple workflows.
> 
> **Overview**
> Introduces a shared `prompt-fragments` module that centralizes common system-prompt text (metadata boundary warning, no-fabrication + structured-output constraints, storyboard reading instructions, tone guidance, confidence rubric, and media-specific language guidelines).
> 
> Updates `ask-questions`, `burned-in-captions`, `engagement-insights`, and `summarization` to interpolate these shared fragments and to build system prompts with a new `promptDedent` (`dedent` with `alignValues: true`) to preserve indentation when inserting multi-line fragments, reducing duplicated prompt text and preventing drift.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f8e052c5c03aa01867e0d4020876dcc0b44fd2f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #235, opened 2026-08-12
fix: Fix missing language guides in ask-questions

philcluff:
## Summary

The `askQuestions` prompts keep producing meta-descriptive reasoning like "The storyboard shows the rabbit setting up traps…" despite guidance to describe content directly.

Turns out ask-questions never got the full anti-meta-description avoid-list — when #174 consolidated shared prompt text into `createLanguageGuidelines()`, only summarization adopted it, and ask-questions kept its original weak three-line block whose only example was "The video shows…".

This swaps both the video and audio-only system prompts over to the shared fragment, which explicitly bans "The storyboard shows…", "The image shows…", "The frames depict…" etc., while keeping the "be specific and evidence-based" instruction.

## Example

Before: "The storyboard shows the rabbit setting up traps that result in the bullies getting caught."
After: "The rabbit sets up traps that catch the bullies by the end of the sequence."

> [!NOTE]
> **Low Risk**
> Prompt-only change that reuses an existing shared fragment; no logic, API, or security-path changes.
> 
> **Overview**
> Restores the fuller anti-meta-description language guidance in `askQuestions` by swapping the short inline examples for the shared `createLanguageGuidelines` helper.
> 
> Both the video and audio-only system prompts now call `createLanguageGuidelines("video")` / `createLanguageGuidelines("audio")`, matching the richer avoid-lists already used in summarization, while keeping the “be specific and evidence-based” lead-in.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fea9bf3c08d09583fb69f3bbafc83d3736fa0105. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/workflows/ask-questions.ts
+++ b/src/workflows/ask-questions.ts
@@ -21,6 +21,7 @@ import { createTranscriptSection, renderSection } from "../lib/prompt-builder.ts
 import {
   CANARY_TRIPWIRE,
   CONFIDENCE_SCORING_RUBRIC,
+  createLanguageGuidelines,
   METADATA_BOUNDARY_WARNING,
   NO_FABRICATION_CONSTRAINT,
   NON_DISCLOSURE_CONSTRAINT,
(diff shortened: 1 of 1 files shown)


monsieurBoutte (approved):
good catch

PR #241, opened 2026-09-01
perf: reuse asset snapshots in workflows

walkerfrankenberg:
# Overview

Allow callers that already retrieved a Mux asset to pass that snapshot into `@mux/ai` workflows. This removes the second Video asset GET that otherwise happens when a workflow resolves playback metadata.

Companion to [muxinc/robots#533](https://github.com/muxinc/robots/pull/533) for [AIT-411](https://mux1.atlassian.net/browse/AIT-411).

# What was changed

- Added an optional `assetSnapshot` field to shared workflow options.
- Updated playback resolution to use the supplied snapshot after validating its asset ID.
- Threaded the snapshot through every workflow that resolves asset playback metadata.
- Added unit coverage for snapshot reuse and mismatched asset IDs.
- Verified with typecheck, lint, build, and all 490 unit tests.

# Suggested review order

1. `src/types.ts` and `src/lib/mux-assets.ts`
2. The workflow call-site updates in `src/workflows/`
3. `tests/unit/mux-assets.test.ts`

### DIFF
--- a/src/lib/mux-assets.ts
+++ b/src/lib/mux-assets.ts
@@ -48,13 +48,17 @@ function toPlaybackAsset(asset: MuxAsset): PlaybackAsset {
 export async function getPlaybackIdForAsset(
   assetId: string,
   credentials?: WorkflowCredentialsInput,
+  assetSnapshot?: MuxAsset,
 ): Promise<PlaybackAsset> {
   "use step";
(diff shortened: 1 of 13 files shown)


electrachong (approved):
This adds the assetSnapshot to the options but there’s still no where we yet provide it in the workflows, right? Is that intended to be a follow-up PR?

PR #246, opened 2026-09-11
feat: generateText tool

philcluff:
# Summary

Adds a `generateText` workflow that writes a set of source-grounded short- and long-form text artifacts (social posts, blog posts, newsletter entries) from a Mux asset. The transcript is the authoritative source. Video assets also attach a scoped storyboard, and can optionally attach a sample of Mux shot frames.

Generation runs in two stages. One editorial brief is extracted from the transcript and images first, then every requested variant × artifact combination is written from that shared brief in parallel. This keeps a whole set coherent while giving each artifact its own length budget, and lets uninstructed variants produce genuinely independent takes.

- `src/workflows/generate-text.ts`: the workflow, its option and result types, option validation, channel length defaults and hard caps, shot-frame sampling, prompts, and the two model steps. Follows the same shape as `generateChapters`: a `"use workflow"` entry wrapping an internal function with `rethrowWithTokenUsage`, model calls inside `"use step"` functions via `createLanguageModelFromConfig` and `withContentPolicyAwareRetry`.
- `artifacts` (1–5, unique keys) are `short_form` with an optional `channel` (`generic`, `x`, `linkedin`, `facebook`, `instagram`, `tiktok`, `youtube`) or `long_form`. Each accepts an optional `maxLength` and bounded `instructions`. `variants` (1–5) default to a single `default` variant.
- `audience`, `voice`, `callToAction`, and `brandTerms` are bounded steering controls rendered into dedicated prompt sections. `scope` bounds transcript cues, the storyboard, and shot selection.
- Length caps are validated before any Mux or model call and enforced after generation with locale-aware word segmentation, so Markdown syntax is not counted and scripts without spaces are measured correctly. A draft that is empty or over a cap gets one corrective rewrite with the measured overshoot fed back; if it still misses, the workflow fails with a retryable `processing_error`. `x` artifacts are always held to 280 characters.
- The editorial brief is scrubbed before fan-out, so a transcript-borne injection is caught after one model call rather than multiplied across the set. Every artifact then passes through the same scrubber; a suppressed artifact returns empty `content` and is named in `result.safety.scrubbedFields`. Unexpected keys emitted by the model are surfaced through the same report.
- Artifacts are written in parallel batches of five with `Promise.allSettled`, and usage from every completed call is preserved on the error when one fails. Steering and per-item instructions are rendered into the user turn, never the system prompt, and language codes must be well-formed BCP 47 tags.
- `useShots` reuses existing shots and requests generation only when the asset has none; `shotPolling` controls how long to wait, defaulting to about five minutes.
- `src/lib/output-safety.ts`: registers the new prompt tag names with the leak detector. Steering tags are prefixed (`steering_voice`, not `voice`) so generic words in legitimate output cannot trip the repo-wide scrubber.
- Exported from `@mux/ai/workflows`, with unit tests, integration and Workflow DevKit integration tests, an example script (`npm run example:generate-text`), and documentation in `README.md`, `docs/WORKFLOWS.md`, `docs/API.md`, and `docs/EXAMPLES.md`.

Not in scope: `promptOverrides`, `imageSubmissionMode`, and provider-specific reasoning controls. All can be added later without changing the result shape.

# Example

```ts
import { generateText } from "@mux/ai/workflows";

const result = await generateText(assetId, {
  variants: [
    { key: "product_led", instructions: "Use a promotional, product-led angle." },
    { key: "insight_led" },
  ],
  artifacts: [
    { key: "x_post", kind: "short_form", channel: "x" },
    { key: "blog_post", kind: "long_form", maxLength: { unit: "words", value: 800 } },
  ],
  audience: "Video developers",
  voice: "conversational",
  callToAction: "soft",
  brandTerms: ["Mux"],
});
```

# Suggested Review order
1. `src/workflows/generate-text.ts`, top to bottom: public types, `resolveGenerateTextOptions`, the length policy, then the prompts. The prompts are where the grounding and non-recap rules live.
2. The two steps (`extractBriefWithModel`, `generateArtifactWithModel`) and `generateTextInternal`, checking the scope, storyboard, shot, usage, and scrubbing paths against `summarization.ts` and `engagement-insights.ts`.
3. `tests/unit/generate-text.test.ts` for the behaviour each path is pinned to.
4. `src/lib/output-safety.ts`, docs, and the example.

---

> [!NOTE]
> **Medium Risk**
> New multi-step LLM workflow with user-controlled instructions and published model output; mitigations mirror other workflows (validation, grounding prompts, output scrubbing) but cost, latency, and content quality still depend on provider behavior.
> 
> **Overview**
> Adds a new **`generateText`** workflow that turns a Mux asset’s transcript (and optional scoped storyboard / shot frames) into source-grounded marketing copy. It first extracts one shared editorial brief, then writes every **variant × artifact** combination in parallel.
> 
> Callers configure **1–5 artifacts** (`short_form` with channel-specific defaults and caps, or `long_form`) and **1–5 variants**, plus optional steering (`audience`, `voice`, `callToAction`, `brandTerms`), `scope`, and `useShots` on video. Length limits are validated up front and enforced after generation (over-cap artifacts fail with a non-retryable `processing_error`). Each artifact is run through the existing output-safety scrubber; suppressed text returns empty `content` with a `safety` report.
> 
> The workflow is exported from `@mux/ai/workflows`, documented in README/API/WORKFLOWS, wired to `npm run example:generate-text`, and covered by unit and integration tests. **`output-safety`** gains prompt-tag names used by the new prompts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd25c85a0c9992117e5f91dc9be8862a7ed58770. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,6 +74,7 @@ Workflows are high-level functions that handle complete media AI tasks end-to-en
 | [`askQuestions`](./docs/WORKFLOWS.md#ask-questions) | Answer yes/no questions about asset content | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
 | [`generateChapters`](./docs/WORKFLOWS.md#chapter-generation) | Create chapter markers from transcripts | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
 | [`generateEmbeddings`](./docs/WORKFLOWS.md#embeddings) | Generate vector embeddings for semantic search | OpenAI, Google, Baseten, OpenAI-compatible | Yes |
+| [`generateText`](./docs/WORKFLOWS.md#text-generation) | Write source-grounded posts and articles from a transcript | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
 | [`translateCaptions`](./docs/WORKFLOWS.md#caption-translation) | Translate captions into other languages | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
 | [`editCaptions`](./docs/WORKFLOWS.md#caption-editing) | Edit captions: censor profanity and/or apply static replacements | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
 | [`translateAudio`](./docs/WORKFLOWS.md#audio-dubbing) | Create AI-dubbed audio tracks | ElevenLabs | Yes |
--- a/docs/API.md
+++ b/docs/API.md
@@ -19,8 +19,8 @@ beginning up to 90 seconds. Times remain relative to the original asset in
 workflow outputs.
 
 `scope` is supported by `getSummaryAndTags`, `getModerationScores`,
-`hasBurnedInCaptions`, `askQuestions`, `generateChapters`, and
-`generateEmbeddings`. It bounds storyboard or thumbnail selection and
+`hasBurnedInCaptions`, `askQuestions`, `generateChapters`,
+`generateEmbeddings`, and `generateText`. It bounds storyboard or thumbnail selection and
 transcript cues as applicable. The workflow rejects negative, non-finite,
 empty, reversed, or out-of-asset ranges.
 
@@ -556,6 +556,56 @@ player.addChapters([
   { startTime: 120, title: "Conclusion" }
 ]);
 ```
+
+## `generateText(assetId, options)`
+
+Writes a set of source-grounded short- and long-form text artifacts from a video or audio asset. Extracts one editorial brief from the transcript (plus a scoped storyboard for video assets), then writes every variant × artifact combination from that brief in parallel.
+
+**Parameters:**
+
+- `assetId` (string) - Mux asset ID (video or audio-only, must have a ready caption track)
+- `options` - Configuration options
+
+**Options:**
+
+- `artifacts: GenerateTextArtifact[]` - Deliverables to write for every variant (1-5, unique lowercase snake_case keys). Each artifact accepts optional `instructions` (up to 500 characters).
+  - `{ key, kind: "short_form", channel?, maxLength?, instructions? }` - `channel` is one of `generic` (default), `x`, `linkedin`, `facebook`, `instagram`, `tiktok`, `youtube`. `maxLength` is `{ unit: "characters", value: 10-5000 }` or `{ unit: "words", value: 5-500 }`; `x` artifacts are capped at 280 characters.
+  - `{ key, kind: "long_form", maxLength?, instructions? }` - `maxLength` is `{ unit: "words", value: 100-3000 }` (default 1200).
+- `variants?: Array<{ key: string; instructions?: string }>` - Named versions of the complete artifact set (1-5, unique keys). Defaults to `[{ key: "default" }]`. Omit `instructions` for an independent take on the same brief.
+- `provider?: 'openai' | 'anthropic' | 'google' | 'baseten' | 'openai-compatible'` - AI provider (default: 'openai')
+- `model?: string` - AI model to use. Defaults per provider: `gpt-5.6-luna` at medium reasoning (OpenAI), `claude-sonnet-4-5` (Anthropic), `gemini-3-flash-preview` (Google). Baseten and OpenAI-compatible endpoints have no default — pass `model` or set the `BASETEN_MODEL` / `OPENAI_COMPATIBLE_MODEL` environment variable.
+- `audience?: string` - Intended reader, used as best-effort guidance (up to 160 characters)
+- `voice?: 'conversational' | 'editorial' | 'playful' | 'professional'` - Best-effort writing voice for every artifact
+- `callToAction?: 'none' | 'soft' | 'direct'` - Whether and how the text should invite the reader to engage further
+- `brandTerms?: string[]` - Brand or domain terms to use exactly when the source supports them (1-10 terms, 40 characters each, 240 combined)
+- `useShots?: boolean` - Attach up to 24 evenly sampled shot frames as extra visual evidence, generating shots if needed (default: false, video assets only)
+- `languageCode?: string` - Language code for transcript track selection. When omitted, prefers English if available.
+- `outputLanguageCode?: string` - BCP 47 language code for the generated text. When omitted or `'auto'`, follows the selected transcript track's language when it is reliable.
+- `scope?: WorkflowScope` - Asset-relative execution window. Bounds transcript cues, the storyboard, and shot selection.
+
+**Returns:**
+
+```typescript
+interface GenerateTextResult {
+  assetId: string;
+  variants: Array<{
+    key: string;
+    artifacts: Array<{
+      key: string;
+      kind: "short_form" | "long_form";
+      content: string; // Empty when suppressed by the output-safety scrubber
+    }>;
+  }>;
+  storyboardUrl?: string; // Storyboard attached to the brief (undefined for audio-only assets)
+  usage?: TokenUsage; // Aggregate usage across the brief and every artifact
+  safety?: SafetyReport; // Output-side scrubbing report
+}
+```
+
+**Errors:**
+
+- `validation_error` - Invalid option bounds, missing transcript, empty scope, or `useShots` on an audio-only asset. Thrown before any model call.
+- `processing_error` - An artifact exceeded its length cap (non-retryable), or no shots overlapped the scope.
 
 ## `translateAudio(assetId, toLanguageCode, options?)`
 
--- a/docs/API.md
+++ b/docs/API.md
@@ -569,7 +569,7 @@ Writes a set of source-grounded short- and long-form text artifacts from a video
 **Options:**
 
 - `artifacts: GenerateTextArtifact[]` - Deliverables to write for every variant (1-5, unique lowercase snake_case keys). Each artifact accepts optional `instructions` (up to 500 characters).
-  - `{ key, kind: "short_form", channel?, maxLength?, instructions? }` - `channel` is one of `generic` (default), `x`, `linkedin`, `facebook`, `instagram`, `tiktok`, `youtube`. `maxLength` is `{ unit: "characters", value: 10-5000 }` or `{ unit: "words", value: 5-500 }`; `x` artifacts are capped at 280 characters.
+  - `{ key, kind: "short_form", channel?, maxLength?, instructions? }` - `channel` is one of `generic` (default), `x`, `linkedin`, `facebook`, `instagram`, `tiktok`, `youtube`. `maxLength` is `{ unit: "characters", value: 10-5000 }` or `{ unit: "words", value: 5-500 }`; `x` artifacts are always held to 280 characters, even when `maxLength` is expressed in words.
   - `{ key, kind: "long_form", maxLength?, instructions? }` - `maxLength` is `{ unit: "words", value: 100-3000 }` (default 1200).
 - `variants?: Array<{ key: string; instructions?: string }>` - Named versions of the complete artifact set (1-5, unique keys). Defaults to `[{ key: "default" }]`. Omit `instructions` for an independent take on the same brief.
 - `provider?: 'openai' | 'anthropic' | 'google' | 'baseten' | 'openai-compatible'` - AI provider (default: 'openai')
--- a/docs/WORKFLOWS.md
+++ b/docs/WORKFLOWS.md
@@ -518,7 +518,7 @@ for (const variant of result.variants) {
 
 - `artifacts` (1-5, unique snake_case keys) are the deliverables. `kind: "short_form"` accepts an optional `channel` (`generic`, `x`, `linkedin`, `facebook`, `instagram`, `tiktok`, `youtube`) whose conventions guide the writing and set a default length cap. `kind: "long_form"` produces developed prose. Either kind accepts `maxLength` as a hard cap and optional bounded `instructions`.
 - `variants` (1-5, unique keys) each receive the complete artifact set. A variant key is only an identifier. Omit `instructions` for an independent take on the same brief, or supply them for a deliberate angle. When `variants` is omitted a single `default` variant is written.
-- Length caps are enforced after generation. An artifact that exceeds its cap fails the workflow with a non-retryable `processing_error`.
+- Length caps are enforced after generation. An artifact that exceeds its cap fails the workflow with a non-retryable `processing_error`. `x` artifacts are additionally held to 280 characters regardless of the unit used for `maxLength`.
 
 | Channel | Default cap |
 | --- | --- |
--- a/docs/WORKFLOWS.md
+++ b/docs/WORKFLOWS.md
@@ -27,7 +27,7 @@ const result = await getSummaryAndTags(assetId, {
 
 Either boundary can be omitted. Scoped execution is available for
 summarization, moderation, burned-in caption detection, question answering,
-chapter generation, and embeddings. Visual workflows request a scoped
+chapter generation, embeddings, and text generation. Visual workflows request a scoped
 storyboard or scoped thumbnails, and transcript-based workflows include only
 cues that overlap the range. Returned timestamps remain relative to the full
 asset.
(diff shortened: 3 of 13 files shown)


docs/API.md
```diff
@@ -557,6 +557,56 @@ player.addChapters([
...
+  - `{ key, kind: "long_form", maxLength?, instructions? }` - `maxLength` is `{ unit: "words", value: 100-3000 }` (default 1200).
+- `variants?: Array<{ key: string; instructions?: string }>` - Named versions of the complete artifact set (1-5, unique keys). Defaults to `[{ key: "default" }]`. Omit `instructions` for an independent take on the same brief.
+- `provider?: 'openai' | 'anthropic' | 'google' | 'baseten' | 'openai-compatible'` - AI provider (default: 'openai')
+- `model?: string` - AI model to use. Defaults per provider: `gpt-5.6-luna` at medium reasoning (OpenAI), `claude-sonnet-4-5` (Anthropic), `gemini-3-flash-preview` (Google). Baseten and OpenAI-compatible endpoints have no default — pass `model` or set the `BASETEN_MODEL` / `OPENAI_COMPATIBLE_MODEL` environment variable.
+- `audience?: string` - Intended reader, used as best-effort guidance (up to 160 characters)
+- `voice?: 'conversational' | 'editorial' | 'playful' | 'professional'` - Best-effort writing voice for every artifact
```
electrachong: ? (question for review does voice constrain to particular set of instructions, should it be more free-from?)

docs/API.md
```diff
@@ -557,6 +557,56 @@ player.addChapters([
...
+- `callToAction?: 'none' | 'soft' | 'direct'` - Whether and how the text should invite the reader to engage further
```
electrachong: [review] look at prompt for cta

docs/API.md
```diff
@@ -557,6 +557,56 @@ player.addChapters([
...
+
+- `assetId` (string) - Mux asset ID (video or audio-only, must have a ready caption track)
+- `options` - Configuration options
+
+**Options:**
+
+- `artifacts: GenerateTextArtifact[]` - Deliverables to write for every variant (1-5, unique lowercase snake_case keys). Each artifact accepts optional `instructions` (up to 500 characters).
+  - `{ key, kind: "short_form", channel?, maxLength?, instructions? }` - `channel` is one of `generic` (default), `x`, `linkedin`, `facebook`, `instagram`, `tiktok`, `youtube`. `maxLength` is `{ unit: "characters", value: 10-5000 }` or `{ unit: "words", value: 5-500 }`; `x` artifacts are always held to 280 characters, even when `maxLength` is expressed in words.
+  - `{ key, kind: "long_form", maxLength?, instructions? }` - `maxLength` is `{ unit: "words", value: 100-3000 }` (default 1200).
+- `variants?: Array<{ key: string; instructions?: string }>` - Named versions of the complete artifact set (1-5, unique keys). Defaults to `[{ key: "default" }]`. Omit `instructions` for an independent take on the same brief.
```
electrachong: ? have test for multiple artifacts same name (default)? probably zod schema takes care?

docs/API.md
```diff
@@ -557,6 +557,56 @@ player.addChapters([
...
+Writes a set of source-grounded short- and long-form text artifacts from a video or audio asset. Extracts one editorial brief from the transcript (plus a scoped storyboard for video assets), then writes every variant × artifact combination from that brief in parallel.
+
+**Parameters:**
+
+- `assetId` (string) - Mux asset ID (video or audio-only, must have a ready caption track)
+- `options` - Configuration options
+
+**Options:**
+
+- `artifacts: GenerateTextArtifact[]` - Deliverables to write for every variant (1-5, unique lowercase snake_case keys). Each artifact accepts optional `instructions` (up to 500 characters).
```
electrachong: ? validation for unique lower case snake_case

docs/API.md
```diff
@@ -557,6 +557,56 @@ player.addChapters([
...
+- `provider?: 'openai' | 'anthropic' | 'google' | 'baseten' | 'openai-compatible'` - AI provider (default: 'openai')
+- `model?: string` - AI model to use. Defaults per provider: `gpt-5.6-luna` at medium reasoning (OpenAI), `claude-sonnet-4-5` (Anthropic), `gemini-3-flash-preview` (Google). Baseten and OpenAI-compatible endpoints have no default — pass `model` or set the `BASETEN_MODEL` / `OPENAI_COMPATIBLE_MODEL` environment variable.
+- `audience?: string` - Intended reader, used as best-effort guidance (up to 160 characters)
+- `voice?: 'conversational' | 'editorial' | 'playful' | 'professional'` - Best-effort writing voice for every artifact
+- `callToAction?: 'none' | 'soft' | 'direct'` - Whether and how the text should invite the reader to engage further
+- `brandTerms?: string[]` - Brand or domain terms to use exactly when the source supports them (1-10 terms, 40 characters each, 240 combined)
+- `useShots?: boolean` - Attach up to 24 evenly sampled shot frames as extra visual evidence, generating shots if needed (default: false, video assets only)
```
electrachong: ? validation for video assets only

docs/API.md
```diff
@@ -557,6 +557,56 @@ player.addChapters([
...
+- `languageCode?: string` - Language code for transcript track selection. When omitted, prefers English if available.
+- `outputLanguageCode?: string` - BCP 47 language code for the generated text. When omitted or `'auto'`, follows the selected transcript track's language when it is reliable.
```
electrachong: validation for outputlanguage code

docs/API.md
```diff
@@ -557,6 +557,56 @@ player.addChapters([
...
+- `provider?: 'openai' | 'anthropic' | 'google' | 'baseten' | 'openai-compatible'` - AI provider (default: 'openai')
+- `model?: string` - AI model to use. Defaults per provider: `gpt-5.6-luna` at medium reasoning (OpenAI), `claude-sonnet-4-5` (Anthropic), `gemini-3-flash-preview` (Google). Baseten and OpenAI-compatible endpoints have no default — pass `model` or set the `BASETEN_MODEL` / `OPENAI_COMPATIBLE_MODEL` environment variable.
+- `audience?: string` - Intended reader, used as best-effort guidance (up to 160 characters)
+- `voice?: 'conversational' | 'editorial' | 'playful' | 'professional'` - Best-effort writing voice for every artifact
+- `callToAction?: 'none' | 'soft' | 'direct'` - Whether and how the text should invite the reader to engage further
+- `brandTerms?: string[]` - Brand or domain terms to use exactly when the source supports them (1-10 terms, 40 characters each, 240 combined)
+- `useShots?: boolean` - Attach up to 24 evenly sampled shot frames as extra visual evidence, generating shots if needed (default: false, video assets only)
+- `languageCode?: string` - Language code for transcript track selection. When omitted, prefers English if available.
```
electrachong, line 582:
```suggestion
- `languageCode?: string` - BCP 47 language code for transcript track selection. When omitted, prefers English if available.
```

docs/WORKFLOWS.md
```diff
@@ -479,6 +479,70 @@ const vttResult = await generateEmbeddings("your-mux-asset-id", {
...
+for (const variant of result.variants) {
+  for (const artifact of variant.artifacts) {
+    console.log(variant.key, artifact.key, artifact.content);
+  }
+}
+```
+
+### Requirements
+
+- Asset must have a ready caption/transcript track. The transcript is the authoritative source for names, claims, and meaning.
```
electrachong: ? validation for ready caption/transcript track

docs/WORKFLOWS.md
```diff
@@ -479,6 +479,70 @@ const vttResult = await generateEmbeddings("your-mux-asset-id", {
...
+- Video assets attach a scoped storyboard to the brief as visual context. Audio-only assets write from the transcript alone.
+
+### Artifacts and Variants
+
+- `artifacts` (1-5, unique snake_case keys) are the deliverables. `kind: "short_form"` accepts an optional `channel` (`generic`, `x`, `linkedin`, `facebook`, `instagram`, `tiktok`, `youtube`) whose conventions guide the writing and set a default length cap. `kind: "long_form"` produces developed prose. Either kind accepts `maxLength` as a hard cap and optional bounded `instructions`.
+- `variants` (1-5, unique keys) each receive the complete artifact set. A variant key is only an identifier. Omit `instructions` for an independent take on the same brief, or supply them for a deliberate angle. When `variants` is omitted a single `default` variant is written.
+- Length caps are enforced after generation. An artifact that exceeds its cap fails the workflow with a non-retryable `processing_error`. `x` artifacts are additionally held to 280 characters regardless of the unit used for `maxLength`.
```
electrachong, line 521: What dictates whether we make the error retryable or not? Reason for keeping failed length cap non-retryable?

src/workflows/generate-text.ts
```diff
@@ -0,0 +1,998 @@
...
+const CHARACTER_LIMIT_RANGE = { min: 10, max: 5000 };
+const SHORT_FORM_WORD_LIMIT_RANGE = { min: 5, max: 500 };
+const LONG_FORM_WORD_LIMIT_RANGE = { min: 100, max: 3000 };
+const X_MAX_CHARACTERS = 280;
+
+function validationError(message: string): MuxAiError {
+  return new MuxAiError(message, { type: "validation_error" });
+}
+
+function assertKeyedItems(
```
electrachong, line 191: do we have other workflows that require a key?

src/workflows/generate-text.ts
```diff
@@ -0,0 +1,998 @@
...
+      if (!trimmed || trimmed.length > GENERATE_TEXT_MAX_INSTRUCTIONS_CHARS) {
+        throw validationError(
+          `${noun} "${item.key}" instructions must be 1-${GENERATE_TEXT_MAX_INSTRUCTIONS_CHARS} characters.`,
+        );
+      }
+    }
+  }
+}
+
+function assertIntegerInRange(
```
electrachong, line 224: do we have other workflows that require an integer in range?

src/workflows/generate-text.ts
```diff
@@ -0,0 +1,998 @@
...
+      throw validationError(`Artifact "${artifact.key}" targets x and supports at most ${X_MAX_CHARACTERS} characters.`);
+    }
+  } else if (artifact.maxLength.unit === "words") {
+    assertIntegerInRange(artifact.maxLength.value, SHORT_FORM_WORD_LIMIT_RANGE, `Artifact "${artifact.key}" maxLength.value`);
+  } else {
+    throw validationError(`Artifact "${artifact.key}" maxLength.unit must be "characters" or "words".`);
+  }
+}
+
+function assertBoundedText(value: string, max: number, label: string): void {
```
electrachong, line 268: do we have other workflows that make similar assertions on output length?

src/workflows/generate-text.ts
electrachong: claude flags engagement-insights.ts:672-673

src/workflows/generate-text.ts
electrachong: > Yes, at least three more hand-rolled variants, each slightly different:
> - ask-questions.ts:851-853 — checks q.question.length > MAX (no trim, so whitespace-only strings under the cap pass).
> - ask-questions.ts:878-880 — same shape for answerOptions.
> - translate-captions.ts:264-282 (validateNeverTranslateTerms) — trims, but splits empty-check and max-check into two separate throws, and adds a </> content check generate-text's version doesn't have.

at some point might be worth standardizing with a common helper but understand if that's something we want to punt on for now

tests/unit/generate-text.test.ts
```diff
@@ -0,0 +1,433 @@
...
+      publicMessage: "Transcript has no usable content in the requested scope.",
+    });
+  });
+});
+
+describe("resolveGenerateTextOptions", () => {
+  const artifacts = [{ key: "post", kind: "short_form" as const }];
+
+  it("rejects duplicate keys, bad key shapes, and too many items", () => {
+    expect(() => resolveGenerateTextOptions({ artifacts: [...artifacts, ...artifacts] })).toThrow("Duplicate artifact key \"post\".");
```
electrachong, line 347: worth adding a test where key is excluded for multiple artifacts or variants and asserting it will raise for duplicate key?

docs/API.md
electrachong, line 579: Should the caller be able to specify what action to invite?

src/workflows/generate-text.ts
electrachong: I think claude demurred but if we do have a common validation module may make sense to move this there assuming we may reuse in the future

src/workflows/generate-text.ts
```diff
@@ -0,0 +1,998 @@
...
+      ${CANARY_TRIPWIRE}
+    </security>
+
+    <artifact_guidance>
+      ${artifactGuidance}
+      Keep the finished text at or below ${args.limits.map(limit => `${limit.value} ${limit.unit}`).join(" and at or below ")}.
+      Unless a section below specifies otherwise:
+      - Write for an informed general audience.
+      - Prefer natural, conversational phrasing over polished corporate language.
+      - Do not force a call to action; close naturally unless the artifact instructions clearly request one.
```
electrachong, line 620: Not sure it matters, but wonder if this guidance is stronger if we insert steeringGuidance.callToAction here instead of repeated below in a different form.

electrachong (approved):
Overall looks good to me, doing a quick pass mostly for learning with some questions for my edification. No blockers on my end

README.md
```diff
@@ -74,6 +74,7 @@ Workflows are high-level functions that handle complete media AI tasks end-to-en
 | [`askQuestions`](./docs/WORKFLOWS.md#ask-questions) | Answer yes/no questions about asset content | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
 | [`generateChapters`](./docs/WORKFLOWS.md#chapter-generation) | Create chapter markers from transcripts | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
 | [`generateEmbeddings`](./docs/WORKFLOWS.md#embeddings) | Generate vector embeddings for semantic search | OpenAI, Google, Baseten, OpenAI-compatible | Yes |
+| [`generateText`](./docs/WORKFLOWS.md#text-generation) | Write source-grounded posts and articles from a transcript | OpenAI, Anthropic, Google, Baseten, OpenAI-compatible | Yes |
```
walkerfrankenberg, line 77: It's more than just a transcript though

src/workflows/generate-text.ts
```diff
@@ -0,0 +1,998 @@
...
+export const GENERATE_TEXT_MAX_VARIANTS = 5;
+export const GENERATE_TEXT_MAX_ARTIFACTS = 5;
+export const GENERATE_TEXT_MAX_INSTRUCTIONS_CHARS = 500;
+export const GENERATE_TEXT_MAX_AUDIENCE_CHARS = 160;
+export const GENERATE_TEXT_MAX_BRAND_TERMS = 10;
+export const GENERATE_TEXT_MAX_BRAND_TERM_CHARS = 40;
+export const GENERATE_TEXT_MAX_BRAND_TERMS_TOTAL_CHARS = 240;
+export const GENERATE_TEXT_MAX_SHOT_FRAMES = 24;
+
+export const GENERATE_TEXT_VOICES = ["conversational", "editorial", "playful", "professional"] as const;
```
walkerfrankenberg, line 58: We use `tone` elsewhere, is it extracted to a helper we can reuse?

src/workflows/generate-text.ts
```diff
@@ -0,0 +1,998 @@
...
+  /** Optional bounded guidance that gives this variant a deliberate angle. */
+  instructions?: string;
+}
+
+interface GenerateTextArtifactBase {
+  /** Lowercase snake_case identifier used to correlate the returned artifact. */
+  key: string;
+  /** Optional bounded guidance specific to this artifact. */
+  instructions?: string;
+}
```
walkerfrankenberg, line 91: Same interface..?

src/workflows/generate-text.ts
```diff
@@ -0,0 +1,998 @@
...
+  channel?: GenerateTextChannel;
+  /** Hard output cap in characters (10-5000) or words (5-500). */
+  maxLength?: GenerateTextLengthLimit;
+}
+
+/** A developed editorial deliverable such as a blog post or newsletter entry. */
+export interface GenerateTextLongFormArtifact extends GenerateTextArtifactBase {
+  kind: "long_form";
+  /** Hard output cap in words (100-3000). */
+  maxLength?: { unit: "words"; value: number };
```
walkerfrankenberg, line 106: Any reason we can't use characters on `long_form` and then move this into the base

walkerfrankenberg (changes requested):
Two high-level questions before digging in too deep to the implementation (since it might change based on the answers to these):
- why do we need `kind`? The only difference between `short_form` and `long_form` is seemingly the optional `channel` preset, it seems like we could always have this present and then the logic for writing would only involve the `maxLength` prop and would simplify the API surface
- Since we're providing `instructions` within each artifact, I'm considering which of the options are actually necessary at the highest level versus which can be folded into `instructions` at the user level. Specifically, thinking about which of these options could differ based on the artifact. The biggest one that sticks out is `callToAction`, since I could imagine this value being different, but could honestly see others of the typically output-steering level ones (audience, voice)

PR #248, opened 2026-09-15
feat: New track replacement behaviour

philcluff:
# Summary

Adds a shared text track replacement module and wires `translateCaptions` and `editCaptions` onto it, so creating a caption track no longer fails (or silently no-ops) when the asset already has a track in the same language or with the same name.

Mux enforces uniqueness on text track `name` (case-insensitive, trimmed) across all statuses, including `preparing` and `errored`. Previously `translateCaptions` swallowed the resulting 400 and returned success with no track, and `editCaptions` dodged it by appending `(edited)` on every run. Design and Mux behaviour verification are in the Robots repo under `apps/api/internal-docs/text-track-replacement.md`.

**New in `src/lib/mux-tracks.ts`** (also exported from the package root):

- `ReplaceExistingTracksPolicy`: `fail | replace_all | replace_generated`. `replace_generated` only removes Mux ASR tracks (`text_source: generated_*`) and refuses if anything uploaded is in the way.
- `planTextTrackReplacement(asset, target, policy)`: pure. Computes the same-language set (primary subtag match, so `en-US` and `eng` equal `en`) and the name-collision set, and returns `clear | blocked | replace`.
- `replaceAndCreateTextTrack(...)`: `"use step"`. Fresh asset fetch, deletes per policy tolerating 404s, creates, retries once on Mux's duplicate-name 400. Returns `created | blocked | create_failed` rather than throwing on known outcomes.
- `passthrough` audit tag on every track we create, defaulting to `{"mux_ai":{"workflow":"..."}}`, overridable and capped at Mux's 255 chars. Audit only; it never drives deletion.

**`translateCaptions`**: `replaceExistingTracks` (default `fail`, checked before any tokens are spent), `trackName`, `trackPassthrough`, and `replacedTracks` on the result. The swallowed create error is gone.

**`editCaptions`**: same options, default `replace_all`. The edited track replaces the source in place under the source's name, language, and `closed_captions`. `fail` keeps the source and requires `trackName`. If the create fails after the source was deleted, the source is restored from the fetched VTT. `deleteOriginalTrack` and `trackNameSuffix` are deprecated but keep their exact old behaviour; mixing them with the new options is a validation error.

**Breaking changes** (pre-1.0):

- `editCaptions` default output is now the source track's name, not `<source> (edited)`, and the source is deleted before the new track is created.
- `translateCaptions` throws on a same-language or same-name conflict instead of returning success without a track.
- Track name suffixes are capitalised: `(Auto-translated)` and `(Auto-dubbed)`.

# Example

```ts
// Replace Mux's auto-generated Spanish captions with a translation, refusing to touch human uploads
const result = await translateCaptions(assetId, enTrackId, "es", {
  provider: "openai",
  replaceExistingTracks: "replace_generated",
});
result.replacedTracks; // [{ id, name: "Spanish CC", languageCode: "es", textSource: "generated_vod", ... }]

// Keep the original and add a clean copy alongside it
await editCaptions(assetId, trackId, {
  replacements: [{ find: "Mucks", replace: "Mux" }],
  replaceExistingTracks: "fail",
  trackName: "English (clean)",
});
```

# Suggested review order

1. `src/lib/mux-tracks.ts`: the policy semantics in `planTextTrackReplacement`, then the step.
2. `tests/unit/mux-tracks.test.ts`: the matching rules as a table of cases.
3. `src/workflows/edit-captions.ts`: legacy vs new path, and the restore-on-failure branch.
4. `src/workflows/translate-captions.ts`: pre-inference fast-fail and the upload block.
5. `tests/unit/text-track-replacement-workflows.test.ts`.
6. `src/index.ts`, docs, example.

---

> [!NOTE]
> **High Risk**
> Breaking workflow defaults and destructive Mux track deletes (with partial-failure restore only on edit-captions) can change production caption assets; behavior is well-tested but callers must opt into new policies explicitly.
> 
> **Overview**
> Introduces a shared **Mux text track replacement** layer (`planTextTrackReplacement`, `replaceAndCreateTextTrack`) and wires **`translateCaptions`** and **`editCaptions`** through it so caption uploads respect Mux’s same-language / same-name uniqueness instead of failing silently or always appending `(edited)`.
> 
> **`translateCaptions`** gains `replaceExistingTracks` (default **`fail`**, checked before any LLM work), optional `trackName` / `trackPassthrough`, and `replacedTracks` on the result; Mux create failures are surfaced instead of logged and ignored.
> 
> **`editCaptions`** defaults to **`replace_all`**: the edited track replaces the source in place (name, language, `closed_captions`), with **source restoration** from the fetched VTT if create fails after deletes. `deleteOriginalTrack` / `trackNameSuffix` remain as a deprecated legacy path. The edit-captions example CLI swaps `--no-delete` for `--keep-original <track-name>`.
> 
> **Breaking (pre-1.0):** `editCaptions` no longer creates `<source> (edited)` by default; `translateCaptions` errors on conflicts when uploading to Mux; auto track suffixes are **`(Auto-translated)`** / **`(Auto-dubbed)`**. Helpers and types are **exported** from the package root. Docs and unit tests cover policies, retries on duplicate-name 400s, and workflow integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ee9bd8ab970d40420e0bcb6633ac6d19d1874abe. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


src/lib/mux-tracks.ts
```diff
@@ -1,7 +1,218 @@
...
+function isNotFoundError(error: unknown): boolean {
+  return typeof error === "object" && error !== null && (error as { status?: unknown }).status === 404;
+}
+
+/** Default audit tag written to `passthrough` on tracks created by `@mux/ai`. */
+export function buildMuxAiTrackPassthrough(workflow: string): string {
+  return JSON.stringify({ mux_ai: { workflow } });
+}
+
+export function validateTrackPassthrough(passthrough: string | undefined): string | undefined {
```
walkerfrankenberg, line 202: So we're creating a passthrough variable and then can potentially fail a job because of it?

src/workflows/translate-audio.ts
```diff
@@ -509,7 +509,7 @@ async function createAudioTrackOnMux(
   const muxClient = await resolveMuxClient(credentials);
   const mux = await muxClient.createClient();
   const languageName = new Intl.DisplayNames(["en"], { type: "language" }).of(languageCode) || languageCode.toUpperCase();
-  const trackName = `${languageName} (auto-dubbed)`;
+  const trackName = `${languageName} (Auto-dubbed)`;
```
walkerfrankenberg, line 512: Why are we not having the same treatment for `translate-audio`? Should we not add `replaceExistingTracks`? This will cause the same name conflicts we're trying to fix

walkerfrankenberg (approved):
Approved to unblock you, but unless there's a reason for us not to include `translate-audio` in this PR, we should probably be doing that for consistency (either in this PR, or a follow-up)

src/lib/mux-tracks.ts
philcluff: So there's two ways that `passthrough` gets set on the track:

1) We auto generate it in @mux/ai, in which case it's guaranteed to be valid

2) If it's set upstream in `trackPassthrough`, and passed into the workflow 

This guard is only to address a consumer of the SDK passing in an invalid passthrough - which in our case will be Robots, so we're safe there, but anyone using @mux/ai standalone could set something too long, so I'd prefer to have an active reject, rather than a trim, or failure dowstream.

src/workflows/translate-audio.ts
philcluff: Yep, good call - will open a stacked PR for translate-audio today.

philcluff: https://github.com/muxinc/ai/pull/254

